PR #1, opened 2019-02-07
Improve Travis build configuration

lonemeow:
This fixes some issues with the Travis CI setup and makes the tests be ran on both JDK8 and JDK11 (currently the Oracle flavor is used).

The full effect of these changes is:
 - default `install` phase is used (runs `./gradlew assemble` to get dependencies)
 - default `script` phase is used (runs `./gradlew check` to build and run tests and checks)
 - `deploy` phase is customized to only run on the JDK8 build (runs `./gradlew publish` to upload artifacts)

Using the `deploy` phase also fixes issue where PR builds would upload artifacts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 language: java
-install: true
+
 jdk:
 - oraclejdk8
+- oraclejdk11
 
 before_cache:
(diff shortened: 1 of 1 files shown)


smrutilal2 (approved):

mmannerm (approved):
We should include openjdk11

PR #7, opened 2019-03-15
Base histogram implementation

smrutilal2:


core/src/main/java/io/ultrabrew/metrics/Histogram.java
```diff
@@ -0,0 +1,17 @@
+package io.ultrabrew.metrics;
+
+import io.ultrabrew.metrics.util.DistributionBucket;
+
+public class Histogram extends Timer {
```
arungupta: Histogram should extend Gauge, not Timer. While most use cases of Histogram will be related to Timer/Latencies, there do exist use cases for non-time based measurements like object sizes, message size, bytes transferred, time elapsed (non-Timer use case) since connections were established, etc.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.Arrays;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
+
+public class BasicHistogramAggregator extends ConcurrentIntTable implements Aggregator {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BasicHistogramAggregator.class);
+  private static final String[] AGG_FIELDS = {"count", "sum", "min", "max"};
+  private static final int[] IDENTITY = {0, 0, Integer.MAX_VALUE, Integer.MIN_VALUE};
```
arungupta: array type needs to be `long[]` because these are statistics (the gauge part of the histogram vs counts within a bucket). since time is tracked in nanos, approximately 2 seconds will reach `Integer.MAX_VALUE` in nanos.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,204 @@
...
+      final int capacity, final int maxCapacity) {
+    super(AGG_FIELDS.length + buckets.getCount(), capacity, maxCapacity);
+    this.metricId = metricId;
+    this.buckets = buckets;
+    this.bucketCount = buckets.getCount();
+  }
+
+  @Override
+  public void apply(String[] tags, long latency, long timestamp) {
+    apply(tags, (int) latency, timestamp);
```
arungupta: do not cast to `int`, as per previous explanation

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  @Override
+  protected void combine(int[] table, long baseOffset, int latency) {
+    add(table, baseOffset, 0, 1);
```
arungupta: if we keep the base array type as `int[]` we need to write a long into 2 `int`s. need to adjust the offsets accordingly.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,66 @@
...
+    private int binarySearch(final int key) {
+
+        int low = 0;
+        int high = buckets.length - 1;
+
+        int minVal = buckets[low];
+        int maxVal = buckets[high];
+
+        if (key >= maxVal) {
+            return high + 1; // align with underflow
```
arungupta: comment should indicate this is the overflow bucket

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public int getBucketIndex(int latency) {
+        return binarySearch(latency);
+    }
+
+    public String[] getBucketNames() {
+        int bucketCount = buckets.length;
+        String[] names = new String[bucketCount + 1];
+        int i = 0;
+        names[i] = UNDERFLOW;
```
arungupta: tracking underflow should be based on user preference to avoid wasting a bucket for uses where it's not relevant. minor optimization.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,66 @@
+package io.ultrabrew.metrics.util;
+
+public class DistributionBucket {
+
+    private static final String UNDERFLOW = "underflow";
+    private static final String OVERFLOW = "overflow";
+
+    private int[] buckets;
+
+    public DistributionBucket(final int[] buckets) {
```
arungupta: need to throw error if only 1 bucket is passed

arungupta: also validate if buckets aren't in increasing order, or if 2 entries have same value

arungupta: or sort the values and eliminate duplicates

arungupta (changes requested):
Please improve as discussed in person

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
```diff
@@ -109,7 +114,8 @@ public void addReporter(final Reporter reporter) {
     reporters.add(reporter);
   }
 
-  private <T extends Metric> T getOrCreate(final String id, final Class<T> klass) {
+  private <T extends Metric> T getOrCreate(final Class<T> klass, final Object... args) {
+    String id = (String) args[0];
```
lonemeow: This is kinda ugly, I'd prefer having a formal parameter and dealing with it in `initArgs` construction.

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
```diff
@@ -120,8 +126,16 @@ public void addReporter(final Reporter reporter) {
...
-        T instance = klass.getDeclaredConstructor(MetricRegistry.class, String.class)
-            .newInstance(this, id);
+        Class<?>[] argTypes = new Class[args.length + 1];
+        Object[] initargs = new Object[args.length + 1];
+        int i = 0;
+        argTypes[i] = MetricRegistry.class;
+        initargs[i] = this;
+        for (; i < args.length; i++) {
+          argTypes[i + 1] = args[i].getClass();
+          initargs[i + 1] = args[i];
```
lonemeow: I would argue this would be more readable as:
```java
argTypes[0] = MetricRegistry.class;
initArgs[0] = this;
for (int i = 1; i < args.length; i++) {
  argTypes[i] = args[i].getClass();
  initArgs[i] = args[i];
```

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
```diff
@@ -120,8 +126,16 @@ public void addReporter(final Reporter reporter) {
         return tryCast(klass, m);
       }
       try {
-        T instance = klass.getDeclaredConstructor(MetricRegistry.class, String.class)
-            .newInstance(this, id);
+        Class<?>[] argTypes = new Class[args.length + 1];
+        Object[] initargs = new Object[args.length + 1];
```
lonemeow: Style nit, correct capitalization to `initArgs`.

core/src/main/java/io/ultrabrew/metrics/Metric.java
```diff
@@ -28,8 +35,34 @@
...
+  }
+
+  /**
+   * Create a metric associated with a metric registry.
+   *
+   * @param registry metric registry the metric is associated with
+   * @param id identifier of the metric
+   * @param cardinality initial cardinality
+   */
+  protected Metric(final MetricRegistry registry, final String id, final int cardinality) {
```
lonemeow: I think this API is the wrong way around, I believe it's far more likely that users are interested in setting the *max* cardinality than the initial one, so that should be easy.

Of course this can be turned around at the public API level but that makes the code harder to read, thus I'd prefer changing this to match the expected cases.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  private Cursor newCursor(boolean sorted) {
+    String[] fields = buildFields();
+    Type[] types = buildTypes(fields.length);
+    long[] identity = buildIdentity(fields.length);
+    return new CursorImpl(tagSets, fields, types, identity, sorted);
+  }
+
+  private long[] buildIdentity(int length) {
```
lonemeow: Since neither the source nor the result array is ever modified, this seems pointless.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+
+  private Type[] buildTypes(final int length) {
+    Type[] types = new Type[length];
+    for (int i = 0; i < length; i++) {
+      types[i] = Type.LONG;
+    }
+    return types;
+  }
+
+  private class CursorImpl implements Cursor {
```
lonemeow, line 158: This class duplicates a lot of code from `ConcurrentMonoidHashTable.CursorImpl`, is there really no way to share the common parts?

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,94 @@
...
+// Licensed under the terms of the Apache License 2.0 license. See LICENSE file in Ultrabrew Metrics
+// for terms.
+
+package io.ultrabrew.metrics.util;
+
+import static io.ultrabrew.metrics.util.Commons.checkArgument;
+
+import java.util.function.Predicate;
+
+public class DistributionBucket {
```
lonemeow: Missing javadoc.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,94 @@
...
+
+    this.buckets = buckets.clone();
+  }
+
+  public int getCount() {
+    return buckets.length + 1; // includes the underflow and overflow buckets
+  }
+
+  public int getBucketIndex(long value) {
+    return binarySearch(value);
```
lonemeow: `Arrays.binarySearch()` instead of custom implementation.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,94 @@
...
+      } else if (key < midVal) {
+        high = mid - 1;
+      } else {
+        low = mid + 1;
+      }
+    }
+    return low;
+  }
+
+  private boolean isSorted(final long[] buckets) {
```
lonemeow: Should be `static`.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,94 @@
...
+    return checkCondition(buckets, i -> buckets[i] < buckets[i + 1]);
+  }
+
+  private boolean hasDuplicate(final long[] buckets) {
```
lonemeow: Should be `static`.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,94 @@
...
+    return checkCondition(buckets, i -> buckets[i] == buckets[i + 1]);
+  }
+
+  private boolean checkCondition(final long[] buckets, Predicate<Integer> predicate) {
```
lonemeow: Should be `static`, rename to something that describes what it does better (`allMatch()` or something like that maybe).

core/src/main/java/io/ultrabrew/metrics/util/Intervals.java
```diff
@@ -9,4 +9,12 @@
   public static long calculateDelay(final long interval, final long currentTime) {
     return interval - (currentTime % interval);
   }
+
+  public static long start() {
```
lonemeow: Given this is only used in tests, move to a class in that scope.

core/src/main/java/io/ultrabrew/metrics/util/Intervals.java
```diff
@@ -9,4 +9,12 @@
...
+    return System.nanoTime();
+  }
+
+  public static long stop(final long startTime) {
```
lonemeow: Given this is only used in tests, move to a class in that scope.

core/src/test/java/io/ultrabrew/metrics/HistogramTest.java
```diff
@@ -0,0 +1,61 @@
...
+import io.ultrabrew.metrics.util.DistributionBucket;
+import io.ultrabrew.metrics.util.Intervals;
+import mockit.Expectations;
+import mockit.Verifications;
+import org.junit.jupiter.api.Test;
+
+public class HistogramTest {
+
+  @Test
+  void testHistoGram() {
```
lonemeow: Given this tests multiple aspects separately, it should be split into sepate test methods for each test.

core/src/test/java/io/ultrabrew/metrics/HistogramTest.java
```diff
@@ -0,0 +1,61 @@
...
+    MetricRegistry metricRegistry = new MetricRegistry();
+    DistributionBucket bucket = new DistributionBucket(new long[]{0, 10, 100});
+    Histogram histogram = metricRegistry.histogram("latency", bucket);
+
+    new Expectations(histogram) {{
+      histogram.emit(anyLong, (String[]) any);
+    }};
+
+    final long startTime = Intervals.start();
+    histogram.set(Intervals.stop(startTime));
```
lonemeow: Using actual time here is completely pointless, just use whatever `long` value you want.

core/src/test/java/io/ultrabrew/metrics/data/BasicCounterAggregatorTest.java
```diff
@@ -491,4 +491,12 @@ public void readingByInvalidTagSets() {
 
     assertFalse(cursor.next());
   }
+
+  @Test
+  void testDefaultCardinality() {
+    BasicCounterAggregator aggregator = new BasicCounterAggregator("test");
+    int maxCapacity = Deencapsulation.getField(aggregator, "maxCapacity");
```
lonemeow, line 501: `Deencapsulation` is being removed from jmockit so it should not be used in new tests, test on observable behaviour instead (in this case, set a max cardinality and check how many items can actually be put in before failure).

core/src/main/java/io/ultrabrew/metrics/Histogram.java
```diff
@@ -0,0 +1,20 @@
+// Copyright 2019, Oath Inc.
+// Licensed under the terms of the Apache License 2.0 license. See LICENSE file in Ultrabrew Metrics
+// for terms.
+
+package io.ultrabrew.metrics;
+
+import io.ultrabrew.metrics.util.DistributionBucket;
+
+public class Histogram extends Gauge {
```
lonemeow: Missing javadoc.

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
```diff
@@ -81,23 +82,27 @@ public GaugeDouble gaugeDouble(final String id) {
    * exists
    */
   public Timer timer(final String id) {
-    return getOrCreate(id, Timer.class);
+    return getOrCreate(Timer.class, id);
+  }
+
+  public Histogram histogram(final String id, DistributionBucket bucket) {
```
lonemeow: Missing javadoc.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+import static io.ultrabrew.metrics.Metric.DEFAULT_MAX_CARDINALITY;
+
+import io.ultrabrew.metrics.Histogram;
+import io.ultrabrew.metrics.util.DistributionBucket;
+import java.util.Arrays;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
+
+public class BasicHistogramAggregator extends ConcurrentIntTable implements Aggregator {
```
lonemeow: Missing javadoc.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  @Override
+  public Cursor sortedCursor() {
+    return newCursor(true);
+  }
+
+  private Cursor newCursor(boolean sorted) {
+    String[] fields = buildFields();
+    Type[] types = buildTypes(fields.length);
```
lonemeow: This should be cached in the aggregator instead of being recreated every time a new cursor is created.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+    return newCursor(false);
+  }
+
+  @Override
+  public Cursor sortedCursor() {
+    return newCursor(true);
+  }
+
+  private Cursor newCursor(boolean sorted) {
+    String[] fields = buildFields();
```
lonemeow: This should be cached in the aggregator instead of being recreated every time a new cursor is created.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+    String[] buckets = this.buckets.getBucketNames();
+    System.arraycopy(AGG_FIELDS, 0, fields, 0, AGG_FIELDS.length);
+    System.arraycopy(buckets, 0, fields, AGG_FIELDS.length, buckets.length);
+    return fields;
+  }
+
+  private Type[] buildTypes(final int length) {
+    Type[] types = new Type[length];
+    for (int i = 0; i < length; i++) {
+      types[i] = Type.LONG;
```
lonemeow: `Arrays.fill(types, Type.LONG)` instead of this loop.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+    return types;
+  }
+
+  private class CursorImpl implements Cursor {
+
+    final private String[] fields;
+    final private Type[] types;
+    private long[] identity;
```
lonemeow: Should be `final` if not removed completely (see comments above).

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentIntTable.java
```diff
@@ -0,0 +1,457 @@
...
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
+
+public abstract class ConcurrentIntTable {
```
lonemeow: This class duplicates a lot of code from `ConcurrentMonoidHashTable`, is there really nothing that can be shared between them (without sacrificing performance obviously)?

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentIntTable.java
```diff
@@ -0,0 +1,457 @@
...
+   */
+  private List<AtomicInteger> recordCounts;
+  /**
+   * Max number of entries allowed in the corresponding table in {@link #tables}.
+   */
+  private List<Integer> tableCapacities;
+  private final int recordSize;
+  private final int maxCapacity;
+  private final long[] identity;
+  private int numAggFields;
```
lonemeow: Should be `final`.

core/src/main/java/io/ultrabrew/metrics/util/Commons.java
```diff
@@ -0,0 +1,14 @@
+// Copyright 2019, Oath Inc.
+// Licensed under the terms of the Apache License 2.0 license. See LICENSE file in Ultrabrew Metrics
+// for terms.
+
+package io.ultrabrew.metrics.util;
+
+public class Commons {
+
+  public static void checkArgument(final boolean condition, final String message) {
```
lonemeow: This method (and/or the class) could use better name, it would be nice to see straight away if it throws if the condition is `true`/`false` without reading code or javadoc.

lonemeow (changes requested):

core/src/main/java/io/ultrabrew/metrics/Histogram.java
SeanPMiller: Monitoring tools such as _statsd_ use "histogram" and "timer" as synonyms. Confusion is inevitable. Whatever the histogram-enabled class is called, it is-a gauge and therefore should extend it.

SeanPMiller: That's a long-winded way of saying I agree but not necessarily for the same reason. Having `Timer` distinct from `Histogram` doesn't make sense to me yet. You want some other type indicating that one or more data sketches (of which a histogram is only one example) are to be applied to measurements of the metric identity.

core/src/main/java/io/ultrabrew/metrics/Histogram.java
SeanPMiller: I wrote this on the first agupta comment I saw, but I question whether this should be modeled this way. Will discuss with you both.

mmannerm: There is no need to have a Histogram. Why is there a histogram in instrumentation. Histograms can be used for Gauge or Timer in the reporting side, but what is histogram really instrumenting? It does not make any sense in instrumentation side.

core/src/main/java/io/ultrabrew/metrics/Metric.java
mmannerm: I agree, initial cardinality is almost irrelevant completely.

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
mmannerm: no need for this

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
```diff
@@ -4,6 +4,7 @@
 
 package io.ultrabrew.metrics;
 
+import io.ultrabrew.metrics.util.DistributionBucket;
```
mmannerm: Unused import?

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,214 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BasicHistogramAggregator.class);
+  private static final String[] AGG_FIELDS = {"count", "sum", "min", "max"};
+  private static final long[] IDENTITY = {0L, 0L, Long.MAX_VALUE, Long.MIN_VALUE};
+
+  private final String metricId;
+  private final DistributionBucket buckets;
+  private final int bucketCount;
+
+  public BasicHistogramAggregator(Histogram histogram) {
```
mmannerm: javadocs everywhere

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: We should create a package protected simplecursorimpl and use that and only create override if really needed.

core/src/main/java/io/ultrabrew/metrics/Histogram.java
arungupta: We need to know the type of data that has to be maintained. ie, histogram vs gauge. if we want histogram a reporter concern, the information that the caller wants a histogram has to somehow be passed down. We could create a different metric/gauge registration api to support that.

core/src/main/java/io/ultrabrew/metrics/util/Commons.java
mmannerm: i.e. "assertTrue"

core/src/test/java/io/ultrabrew/metrics/data/BasicCounterAggregatorTest.java
mmannerm: it is already gone, upgrade to latest jmockit as well on the PR.

mmannerm (changes requested):

core/src/main/java/io/ultrabrew/metrics/Histogram.java
mmannerm: There are two distinct phases as per our readme and documentation, etc. One is to instrument code, i.e. I am measuring now this and that etc. How we then report it is a separate issue. I can use histograms, latency buckets, ngrams etc to report on the timer. I must not need to change the instrumentation code when I decide to switch this.

arungupta: Histograms, ngrams, etc require configuration which may be specific to a metric. In case of histograms, that's the bucket spec. Even if we treat Histograms as a reporter level concern, the instrumented code will have to pass the appropriate configurations to the reporter(s) separately from how metrics are registered.

One could argue that metrics should all be predefined in a single place and instrumented code should just use these metrics. I don't believe that'll be feasible for all applications.

We could go for code like this:

```
MetricRegistry metricRegistry = new MetricRegistry();

Reporter metricsReporter = new Reporter(...);
metricRegistry.addReporter(reporter);

Gauge cacheSizeGauge = metricRegistry.gauge("cacheSize");
Map<String,String> cache = new java.util.Map<>();
String[] tagList = new String[] { "host", hostName };
metricsReporter.addAggregator(cacheSizeGauge, Reporter.TYPE_HISTOGRAM, <spec>)

// ... do something

cacheSizeGauge.set(cache.size(), tagList);
```

Thoughts?

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
arungupta: This is a draft PR. Javadocs will follow. @smrutilal2 was trying to get feedback on whatever he had done so far while he's working on remaining parts.

core/src/main/java/io/ultrabrew/metrics/Histogram.java
mmannerm: Resolve conversation button really sucks.

The configuration is reporter specific anyway, you cant give the configuration in the instrumentation class, as you need to support multiple reporters and one might report it as p99 only and other as avg and 3rd as latency buckets. 

There was already ask in our discussion with smruti the need for the config overrides on class AND metric id level

arungupta: Are you thinking of providing overrides in a different class, away from where metric is originally registered? I'm not sure if that's the right approach. Configuration around a metric should be managed in a common place.. preferably where the metric is declared.

mmannerm: That is against the whole design where instrumentation and reporting are separated

arungupta: They would still be separated. Specifically, this configuration would be away from where the metric is set/modified.

mmannerm: No, it should be set where reporter is defined.

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
smrutilal2: done

core/src/main/java/io/ultrabrew/metrics/MetricRegistry.java
smrutilal2: I guess, you are suggesting 
`private <T extends Metric> T getOrCreate(final Class<T> klass, final String id,
      final Object... args)`

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
smrutilal2: The point is not clear. Could you please explain your intent?

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
smrutilal2: This is a custom implementation where the entries in the array represents a range instead of an absolute value. Can't use `Arrays.binarySearch()`. It will be clear when I add the java doc.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentIntTable.java
```diff
@@ -78,8 +76,8 @@ public ConcurrentIntTable(final int recordSize, int initialCapacity, final int m
 
   private ConcurrentIntTable(final int numAggFields, final int dataSize, int initialCapacity, final int maxCapacity, final long[] identity){
 
-    checkArgument(initialCapacity >= 0, "Illegal initial capacity");
-    checkArgument(maxCapacity >= initialCapacity, "max capacity should be greater than the initial capacity");
+    assert initialCapacity >= 0 : "Illegal initial capacity";
```
arungupta: how about renaming this class to `ConcurrentMonoidIntTable` and the other one to `ConcurrentMonoidLongTable`?

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -54,7 +65,8 @@ public SLF4JReporter(final String name) {
    * @param windowSizeSeconds window size in seconds
    */
   public SLF4JReporter(final String name, final int windowSizeSeconds) {
-    this(name, " ", " ", " ", windowSizeSeconds);
+    this(name, DEFULT_TAG_DELIMITER, DEFULT_FIELD_DELIMITER, DEFULT_TAGFIELD_DELIMITER,
```
arungupta: typo `DEFAULT_`

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -13,6 +13,31 @@
...
+ * <li>sum of the measurement values</li>
+ * <li>minimum measured value</li>
+ * <li>maximum measured value</li>
+ * <li>last measured value</li>
+ * </ul>
+ *
+ * The histogram buckets are derived from {@link DistributionBucket}.
+ * <p>A sample histogram would look like:</p>
+ * <ul>
+ * <li>[0_10)</li>
```
arungupta, line 31: While the buckets in this doc look pretty, we are likely going to have an ugly outcome with bucket naming for timer metrics because the unit of measurement is in nanoseconds while most humans think/expect in milliseconds.
Ideally, when the reporters are built, we should allow users to specify the output bucket spec unit so the devs can state they wish to see bucket names in milliseconds even though internally the library tracks them in nanoseconds.

Additionally, we should have an option to prettify the bucket names to avoid things like `52.4498324_58.230823832`, rather `52.5_58.2`, etc

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -41,8 +41,8 @@
 public class BasicHistogramAggregator extends ConcurrentIntTable implements Aggregator {
```
arungupta: as discussed in person with @smrutilal2, we should not absorb the functionality of gauge/timer into histogram aggregator, rather let them run in parallel. this will allow users to eventually have multiple aggregators depending on use cases. eg, timer+histogram+t-digest+HLL for a metric but different downstream use cases.

arungupta (dismissed):

smrutilal2:
@lonemeow @mmannerm could you please review.

core/src/main/java/io/ultrabrew/metrics/data/BasicCounterAggregator.java
```diff
@@ -16,43 +19,50 @@
...
   /**
    * Create a monoid for common aggregation functions for a Counter with requested capacity.
    *
    * @param metricId identifier of the metric associated with this aggregator
-   * @param capacity requested capacity of table in records, actual capacity may be higher
+   * @param cardinality requested capacity of table in records, actual capacity may be higher
    */
-  public BasicCounterAggregator(final String metricId, final int capacity) {
-    super(metricId, capacity, FIELDS, TYPES, IDENTITY);
+  public BasicCounterAggregator(final String metricId, final int cardinality) {
```
mmannerm: I would still make this method `public BasicCounterAggregator(final String metricId, final int maxCardinality)` instead and calculate the initial capacity from the maxCardinality or use sane default.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: what is the point of this method?

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -54,7 +63,8 @@ public SLF4JReporter(final String name) {
    * @param windowSizeSeconds window size in seconds
    */
   public SLF4JReporter(final String name, final int windowSizeSeconds) {
-    this(name, " ", " ", " ", windowSizeSeconds);
+    this(name, DEFAULT_TAG_DELIMITER, DEFAULT_FIELD_DELIMITER, DEFAULT_TAGFIELD_DELIMITER,
```
mmannerm: I really dislike this as now I have to scroll somewhere and find out what the hell are the values here.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -32,6 +37,10 @@
  */
 public class SLF4JReporter extends TimeWindowReporter {
```
mmannerm, line 38: Every change on this class looks like it could be in a completely separate PR, any reason why thats not the case?

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
smrutilal2: I extracted these constants to make it configurable through the Builder. If we don't want to make them configurable then I can inline them.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
smrutilal2: This change is to demonstrate how the reporters are going to use the HistogramAggregator.

core/src/main/java/io/ultrabrew/metrics/util/DistributionBucket.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private static boolean isSorted(final long[] buckets) {
+    return !matchAny(buckets, i -> buckets[i] > buckets[i + 1]);
+  }
+
+  private static boolean hasDuplicate(final long[] buckets) {
+    return matchAny(buckets, i -> buckets[i] == buckets[i + 1]);
+  }
+
+  private static boolean matchAny(final long[] buckets, Predicate<Integer> predicate) {
```
mmannerm: While seems this is only called at constructor time, this will box/unbox the integers and create lots of garbage.. It may be fine here, a javadoc would be good..

core/src/main/java/io/ultrabrew/metrics/util/Intervals.java
```diff
@@ -8,5 +8,6 @@
 
   public static long calculateDelay(final long interval, final long currentTime) {
     return interval - (currentTime % interval);
+
```
mmannerm: remove emptyline

core/src/test/java/io/ultrabrew/metrics/data/BasicCounterAggregatorTest.java
mmannerm: Remove Deencapsulation usage

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
smrutilal2: This is the identity values for the buckets.

core/src/test/java/io/ultrabrew/metrics/data/BasicGaugeAggregatorTest.java
```diff
@@ -112,4 +113,12 @@ public void testGrowTableWithMaxCapacity() {
     assertEquals(3, aggregator.size());
     assertEquals(3, aggregator.capacity()); // caped at the max capacity.
   }
+
+  @Test
+  void testDefaultCardinality() {
+    BasicGaugeAggregator aggregator = new BasicGaugeAggregator("test");
+    int maxCapacity = Deencapsulation.getField(aggregator, "maxCapacity");
```
mmannerm, line 121: Remove Deencapsulation usage

smrutilal2: We will address them as part of https://github.com/ultrabrew/metrics/issues/12

mmannerm (changes requested):
How are the developer supposed to supply the DistributionBuckets ? Example code? At the moment it seems too complicated for average engineers.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: Yes, but why do you need this copy here? Its not modified anywhere?

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
mmannerm: ? I am confused, how does replacing `static final DEFAULT_TAG_DELIMITER = " "` with inline `" "` change anything except readability?

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
mmannerm: Hmmh, can't see it anywhere? Usage of HistogramAggregator is not in default aggregators or anywhere?

smrutilal2: You can see the usage in `SLF4JReporterTest::testHistogram()`

smrutilal2:
> How are the developer supposed to supply the DistributionBuckets ? Example code? At the moment it seems too complicated for average engineers.

You can see the usage in `SLF4JReporterTest::testHistogram`. Client can supply the distribution buckets while creating the Reporter builder. For example:

```Java
DistributionBucket bucket = new DistributionBucket(new long[]{0, 10, 50, 100});
InfluxDBReporter r = InfluxDBReporter.builder()
        .withBaseUri(TEST_URI)
        .withDatabase("test")
        .withWindowSize(12765)
        .addHistogram("requestTime", bucket)
        .build();
```

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
smrutilal2: I pulled out a constant to use it in the builder. If you really hate this then I will have to inline them in two places.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
smrutilal2: These are the identity values used by the Cursor to reset the values after reading. The record length depends on the number of buckets used for the histogram, I initialize this array once, and use it each time the Cursor object is created.

smrutilal2: Currently the Base aggregation like `count`, `sum`,`min`,`max` and buckets counts are tracked in the same table. Once we split that as part of https://github.com/ultrabrew/metrics/issues/18, we can default the identity values for the buckets.

mmannerm (dismissed):

mmannerm:
I still want a better example outside of tests.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,266 @@
...
+    System.arraycopy(IDENTITY, 0, identity, 0, IDENTITY.length);
+    return identity;
+  }
+
+  /**
+   * Creates an array of names of the aggregation field and histogram buckets.
+   */
+  private String[] buildFields() {
+    String[] fields = new String[AGG_FIELDS.length + bucketCount];
+    String[] buckets = this.buckets.getBucketNames();
```
lonemeow: Shadowing members with locals (and then being forced to use `this`) is ugly, please rename this local variable to `bucketNames` or something like that.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,266 @@
...
+ * <li>overflow</li>
+ * <li>underflow</li>
+ * </ul>
+ *
+ * @see DistributionBucket
+ */
+public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implements Aggregator {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(BasicHistogramAggregator.class);
+  private static final String[] AGG_FIELDS = {"count", "sum", "min", "max", "lastValue"};
```
lonemeow: Style nit: What does `AGG` stand for? Avoid ambiguous shorthands.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,266 @@
...
+   * @param buckets distribution bucket spec
+   * @param maxCardinality requested max capacity of table in records. Table doesn't grow beyond
+   * @param cardinality requested capacity of table in records, actual capacity may be higher
+   */
+  public BasicHistogramAggregator(final String metricId, final DistributionBucket buckets,
+      final int maxCardinality, final int cardinality) {
+    super(AGG_FIELDS.length + buckets.getCount(), maxCardinality, cardinality, IDENTITY);
+    this.metricId = metricId;
+    this.buckets = buckets;
+    this.bucketCount = buckets.getCount();
```
lonemeow: Is this field just an optimization for `buckets.getCount()`? If so, is the performance impact sufficient to justify the extra complexity?

core/src/main/java/io/ultrabrew/metrics/data/DistributionBucket.java
```diff
@@ -0,0 +1,138 @@
...
+    return names;
+  }
+
+  /**
+   * retrieves the corresponding bucket index for a given measurement value.
+   *
+   * @param key measurement value
+   * @return bucket index
+   */
+  private int binarySearch(final long key) {
```
lonemeow: I would probably just inline this into `getBucketIndex()` as the only thing it does is call this.

Also a comment explaining what's special about this particular binary search implementation (ie. how it is different from `Arrays.binarySearch()`) should be added.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -142,4 +193,52 @@ private String formatFields(final CursorEntry cursor) {
     }
     return sb == null ? "" : sb.toString();
   }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static final class Builder extends TimeWindowReporterBuilder<Builder, SLF4JReporter> {
```
lonemeow, line 149: Missing javadoc for all methods in this class. Even though they're fairly self descriptive, at least the default values should be documented for each of the optional ones.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -70,7 +80,48 @@ public SLF4JReporter(final String name, final CharSequence tagDelimiter,
...
+   * Create a SLF4J Logger reporter with given delimiters.
+   *
+   * @param name name of the logger
+   * @param tagDelimiter delimiter to be used to join tag key-value pairs
+   * @param fieldDelimiter delimiter to be used to join field name-value pairs
+   * @param tagFieldDelimiter delimiter to be used to separate tags and fields
+   * @param windowSizeSeconds window size in seconds
+   * @param defaultAggregators a map of a metric class to a supplier creating a new aggregator
+   */
+  public SLF4JReporter(final String name, final CharSequence tagDelimiter,
```
lonemeow: No new public constructors should be added if builder is the proposed new API.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -70,7 +80,48 @@ public SLF4JReporter(final String name, final CharSequence tagDelimiter,
...
+   * @param name name of the logger
+   * @param tagDelimiter delimiter to be used to join tag key-value pairs
+   * @param fieldDelimiter delimiter to be used to join field name-value pairs
+   * @param tagFieldDelimiter delimiter to be used to separate tags and fields
+   * @param windowSizeSeconds window size in seconds
+   * @param defaultAggregators a map of a metric class to a supplier creating a new aggregator
+   * @param metricAggregators a map of a metric identifier to a supplier creating a new aggregator
+   * instance
+   */
+  public SLF4JReporter(final String name, final CharSequence tagDelimiter,
```
lonemeow: No new public constructors should be added if builder is the proposed new API.

core/src/main/java/io/ultrabrew/metrics/reporters/SLF4JReporter.java
```diff
@@ -142,4 +193,52 @@ private String formatFields(final CursorEntry cursor) {
...
+    }
+
+    public Builder withTagFieldDelimiter(CharSequence tagFieldDelimiter) {
+      this.tagFieldDelimiter = tagFieldDelimiter;
+      return this;
+    }
+
+    @Override
+    public SLF4JReporter build() {
+      return new SLF4JReporter(name, tagDelimiter, fieldDelimiter, tagFieldDelimiter,
```
lonemeow, line 215: Missing validation of values; at least a `null` check for `name` is required.

core/src/main/java/io/ultrabrew/metrics/reporters/TimeWindowReporter.java
```diff
@@ -41,12 +48,18 @@ public TimeWindowReporter(final String name, final int windowStepSizeSeconds) {
   }
 
   public TimeWindowReporter(final String name, final int windowStepSizeSeconds,
-      final Map<Class<? extends Metric>, Function<String, ? extends Aggregator>> aggregators) {
+      final Map<Class<? extends Metric>, Function<Metric, ? extends Aggregator>> defaultAggregators) {
+    this(name, windowStepSizeSeconds, defaultAggregators, Collections.EMPTY_MAP);
```
lonemeow, line 53: Style nit: Prefer `Collections.emptyMap()` for type correctness.

core/src/main/java/io/ultrabrew/metrics/Metric.java
```diff
@@ -10,15 +10,30 @@
...
+ * a few (5+) fields. The combined time-series cardinality in a JVM can be more than a million (1,000,000). The
+ * default max cardinality of a metric is 4k (4096) beyond which, any new dimension will be dropped.
+ * Clients can pass a custom value to create metric with higher cardinality.
+ *
  * @see MetricRegistry
  */
 public abstract class Metric {
 
+  public static final int DEFAULT_CARDINALITY = 128;
+  public static final int DEFAULT_MAX_CARDINALITY = 4096; //4k
```
lonemeow, line 23: Package protected maybe.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -120,40 +118,19 @@
...
-          "Fields and Identity must match in length and be non-zero");
-    }
-    if (initialCapacity < 0) {
-      throw new IllegalArgumentException("Illegal initial capacity");
-    }
-    if (maxCapacity < initialCapacity) {
-      throw new IllegalArgumentException(
-          "max capacity should be greater than the initial capacity");
-    }
+    assert fields.length != 0 && fields.length == identity.length : "Fields and Identity must match in length and be non-zero";
```
lonemeow: Please don't use `assert` in non-performance-critical code, almost no one ever runs JVM with assertions enabled.

core/src/main/java/io/ultrabrew/metrics/reporters/TimeWindowReporter.java
```diff
@@ -146,4 +159,43 @@ private int getWriterIndex(final long milliseconds) {
...
+
+  /**
+   * A base class for the reporter builder
+   *
+   * @param <B> builder
+   * @param <R> reporter
+   */
+  public static abstract class TimeWindowReporterBuilder<B extends TimeWindowReporterBuilder, R extends TimeWindowReporter> {
+
+    protected Map<Class<? extends Metric>, Function<Metric, ? extends Aggregator>> defaultAggregators = DEFAULT_AGGREGATORS;
```
lonemeow: Awfully long lines here, check formatting and max line length.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
lonemeow: So do I understand this correctly: This method is intended to produce an array that contains copy of `IDENTITY` followed by `buckets.getCount()` zeroes?

If that is the case, please make it explicit in the code instead of relying on default initialized values. A comment explaining why would also help.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -0,0 +1,266 @@
...
+    Type[] types = new Type[length];
+    Arrays.fill(types, Type.LONG);
+    return types;
+  }
+
+  private class CursorImpl implements Cursor {
+
+    private final String[] fields;
+    private final Type[] types;
+    private final long[] identity;
```
lonemeow: Since this is a non-static inner class, `fields`, `types` and `identity` can be accessed directly from the outer class, no need to have duplicate variables here.

core/src/main/java/io/ultrabrew/metrics/data/DistributionBucket.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Creates a distribution for given bucket spec.
+   *
+   * @param buckets sorted array of unique values
+   */
+  public DistributionBucket(final long[] buckets) {
+
+    assert buckets.length >= 2 : "Minimum bucket length is 2";
+    assert isSorted(buckets) : "Bucket should be sorted in ascending order";
+    assert !hasDuplicate(buckets) : "Bucket should not have duplicate entries";
```
lonemeow: Please don't use `assert` in non-performance-critical code, almost no one ever runs JVM with assertions enabled.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
```diff
@@ -0,0 +1,469 @@
...
+  protected ConcurrentMonoidIntTable(final int recordSize, final int maxCapacity, int initialCapacity,
+      final long[] identity) {
+    this(identity.length, recordSize - identity.length, maxCapacity, initialCapacity, identity);
+  }
+
+  private ConcurrentMonoidIntTable(final int numAggFields, final int dataSize,
+      final int maxCapacity, int initialCapacity, final long[] identity) {
+
+    assert initialCapacity >= 0 : "Illegal initial capacity";
+    assert maxCapacity >= initialCapacity : "max capacity should be greater than the initial capacity";
```
lonemeow: Please don't use `assert` in non-performance-critical code, almost no one ever runs JVM with assertions enabled.

lonemeow (changes requested):

lonemeow (approved):

arungupta (approved):

PR #11, opened 2019-03-26
Fix release deployments

lonemeow:
The defaults for Travis only allow deploying from master branch, thus release tag builds would not be deployed.


mmannerm (approved):

mmannerm:
How about forcing it to tags on master, or can’t be done?

smrutilal2 (approved):

Issue #26, opened 2019-12-03
Purge Stale metrics


smrutilal2:
The library need the ability to purge the stale metrics for which it's not recording the values anymore. Without this, the metric will eventually reach the max limit and drop the new cardinality.

PR #28, opened 2019-12-06
First cut: Free up stale rows feature for clients

gskchaitanya:

This PR addresses https://github.com/ultrabrew/metrics/issues/26. The main idea is to let the clients take a call on purging a tagset state that is no longer needed. I've added a method `freeCurrentRow` to the `CursorEntry` so the clients can purge the current row while iterating to report the data. Piggybacked on the tagset itself when a current row is marked for deletion by setting the first tag field to null.

The currently sorted cursor iterator works on a copy of the tag sets and this PR changes that to work on the same tag sets avoiding additional memory space.
Would like to tackle the Histogram implementations in the next phase once we have agreed on the current implementation.

### DIFF
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -161,13 +161,18 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
     private int i = -1;
     private long base = 0;
     private int[] table;
+    private final Integer[] ref;
 
     private CursorImpl(final String[][] tagSets, final boolean sorted) {
+      // Holds the indexes of tagSets in the case of sorted iteration
+      ref = new Integer[tagSets.length];
+      for (int i = 0; i < tagSets.length; i++) {
+        ref[i] = i;
+      }
+      this.tagSets = tagSets;
       if (sorted) {
-        this.tagSets = tagSets.clone();
-        Arrays.sort(this.tagSets, TagSetsHelper::compare);
-      } else {
-        this.tagSets = tagSets;
+        TagSetsHelper comparator = new TagSetsHelper(tagSets); 
+        Arrays.sort(ref, comparator);
       }
     }
 
@@ -179,15 +184,15 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
     @Override
     public boolean next() {
       i++;
-      if (i >= tagSets.length || tagSets[i] == null) {
+      if (i >= tagSets.length || tagSets[ref[i]] == null) {
         return false;
       }
 
-      long index = index(tagSets[i], true);
+      long index = index(tagSets[ref[i]], true);
 
       if (NOT_FOUND == index) {
         LOGGER.error("Missing index on Read. Tags: {}. Concurrency error or bug",
-            Arrays.asList(tagSets[i]));
+            Arrays.asList(tagSets[ref[i]]));
         return false;
       }
 
@@ -204,10 +209,10 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
 
     @Override
     public String[] getTags() {
-      if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + i);
+      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
       }
-      return tagSets[i];
+      return tagSets[ref[i]];
     }
  ///CLOVER:OFF
     @Override
@@ -217,16 +222,16 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
  ///CLOVER:ON
     @Override
     public long lastUpdated() {
-      if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + i);
+      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
       }
       return readTime(table, base);
     }
 
     @Override
     public long readLong(final int index) {
-      if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + i);
+      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
       }
       if (index < 0 || index >= fields.length) {
         throw new IndexOutOfBoundsException("Not a valid field index: " + index);
@@ -241,8 +246,8 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
 
     @Override
     public long readAndResetLong(final int index) {
-      if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + i);
+      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
       }
       if (index < 0 || index >= fields.length) {
         throw new IndexOutOfBoundsException("Not a valid field index: " + index);
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -209,6 +209,12 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
       }
       return tagSets[i];
     }
+    
+    @Override
+    public void freeCurrentRow() {
+      // TODO Auto-generated method stub
+      
+    }
 
     @Override
     public long lastUpdated() {
@@ -259,5 +265,6 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
     public Type[] getTypes() {
       return types;
     }
+
   }
 }
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -209,13 +209,12 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
       }
       return tagSets[i];
     }
-    
+ ///CLOVER:OFF
     @Override
-    public void freeCurrentRow() {
-      // TODO Auto-generated method stub
-      
+    public String[] freeCurrentRow() {
+      return null;
     }
-
+ ///CLOVER:ON
     @Override
     public long lastUpdated() {
       if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -171,7 +171,7 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
       }
       this.tagSets = tagSets;
       if (sorted) {
-        TagSetsHelper comparator = new TagSetsHelper(tagSets); 
+        TagSetsComparator comparator = new TagSetsComparator(tagSets); 
         Arrays.sort(sortedRef, comparator);
       }
     }
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -161,18 +161,18 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
     private int i = -1;
     private long base = 0;
     private int[] table;
-    private final Integer[] ref;
+    private final Integer[] sortedRef;
 
     private CursorImpl(final String[][] tagSets, final boolean sorted) {
       // Holds the indexes of tagSets in the case of sorted iteration
-      ref = new Integer[tagSets.length];
+      sortedRef = new Integer[tagSets.length];
       for (int i = 0; i < tagSets.length; i++) {
-        ref[i] = i;
+        sortedRef[i] = i;
       }
       this.tagSets = tagSets;
       if (sorted) {
         TagSetsHelper comparator = new TagSetsHelper(tagSets); 
-        Arrays.sort(ref, comparator);
+        Arrays.sort(sortedRef, comparator);
       }
     }
 
@@ -184,15 +184,15 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
     @Override
     public boolean next() {
       i++;
-      if (i >= tagSets.length || tagSets[ref[i]] == null) {
+      if (i >= tagSets.length || tagSets[sortedRef[i]] == null) {
         return false;
       }
 
-      long index = index(tagSets[ref[i]], true);
+      long index = index(tagSets[sortedRef[i]], true);
 
       if (NOT_FOUND == index) {
         LOGGER.error("Missing index on Read. Tags: {}. Concurrency error or bug",
-            Arrays.asList(tagSets[ref[i]]));
+            Arrays.asList(tagSets[sortedRef[i]]));
         return false;
       }
 
@@ -209,12 +209,13 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
 
     @Override
     public String[] getTags() {
-      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
+      if (i < 0 || i >= tagSets.length || tagSets[sortedRef[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + sortedRef[i]);
       }
-      return tagSets[ref[i]];
+      return tagSets[sortedRef[i]];
     }
  ///CLOVER:OFF
+    // Will turn on Clover once this method is implemented.
     @Override
     public String[] freeCurrentRow() {
       return null;
@@ -222,16 +223,16 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
  ///CLOVER:ON
     @Override
     public long lastUpdated() {
-      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
+      if (i < 0 || i >= tagSets.length || tagSets[sortedRef[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + sortedRef[i]);
       }
       return readTime(table, base);
     }
 
     @Override
     public long readLong(final int index) {
-      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
+      if (i < 0 || i >= tagSets.length || tagSets[sortedRef[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + sortedRef[i]);
       }
       if (index < 0 || index >= fields.length) {
         throw new IndexOutOfBoundsException("Not a valid field index: " + index);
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -161,13 +161,18 @@ private Cursor newCursor(boolean sorted) {
     private int i = -1;
     private long base = 0;
     private int[] table;
+    private final Integer[] ref;
```
lonemeow: A more descriptive variable name would be nice.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -204,24 +209,29 @@ public boolean next() {
...
-      if (i < 0 || i >= tagSets.length || tagSets[i] == null) {
-        throw new IndexOutOfBoundsException("Not a valid row index: " + i);
+      if (i < 0 || i >= tagSets.length || tagSets[ref[i]] == null) {
+        throw new IndexOutOfBoundsException("Not a valid row index: " + ref[i]);
       }
-      return tagSets[i];
+      return tagSets[ref[i]];
     }
-
+ ///CLOVER:OFF
```
lonemeow, line 217: Justification for this required, why is this untestable?

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -161,13 +161,18 @@ private Cursor newCursor(boolean sorted) {
     private int i = -1;
     private long base = 0;
     private int[] table;
+    private final Integer[] ref;
 
     private CursorImpl(final String[][] tagSets, final boolean sorted) {
+      // Holds the indexes of tagSets in the case of sorted iteration
+      ref = new Integer[tagSets.length];
```
lonemeow: Maybe `Arrays.copyOf()`?

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -294,11 +296,30 @@ long index(final String[] tags, final boolean isReading) {
                 // grow tag set
                 synchronized (this) {
                   if (tagIndex >= tagSets.length) {
+                    boolean isEqual = false;
+                    if(tagIndex == tagSets.length) {
```
lonemeow: `boolean isEqual = tagIndex == tagSets.length` would be less verbose.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -294,11 +296,30 @@ long index(final String[] tags, final boolean isReading) {
...
                         oldLength > TAGSETS_MAX_INCREMENT ? oldLength + TAGSETS_MAX_INCREMENT
                             : oldLength * 2;
-                    tagSets = Arrays.copyOf(tagSets, newLength);
+                    String[][] newTagSet = new String[newLength][];
+                    int cnt = 0;
+                    for (int i = 0; i < tagSets.length; i++) {
+                      // Skip the entries that would have been freed up (stale)
+                      // System.arraycopy would be faster, but can't use it here due to 
+                      // the conditional inclusion 
+                      if ((tagSets[i].length > 0 && tagSets[i][0] != null)
```
lonemeow: Assign `tagSets[i]` to a local variable maybe?

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -294,11 +296,30 @@ long index(final String[] tags, final boolean isReading) {
...
+                          || (tagSets[i].length == 0)) {
+                        String[] newTag = tagSets[i];
+                        newTagSet[cnt++] = newTag;
+                      }
+                    }
+                    tagSets = newTagSet;
+                    // reset the usedOffset length;
+                    unsafe.putIntVolatile(this, usedOffset, isEqual ? cnt + 1 : cnt + 2);
```
lonemeow, line 319: There's a race here between `tagSets` being replaced and `usedOffset` getting rewritten - that seems like a problem unless something elsewhere makes this safe?

core/src/main/java/io/ultrabrew/metrics/data/TagSetsHelper.java
```diff
@@ -4,7 +4,10 @@
 
 package io.ultrabrew.metrics.data;
 
-class TagSetsHelper {
+import java.io.Serializable;
+import java.util.Comparator;
+
+class TagSetsHelper implements Comparator<Integer>, Serializable {
```
lonemeow: Probably should rename the class to `TagSetComparator`.

Why does this implement `Serializable`?

gskchaitanya: It is generally a good idea for comparators to also implement Serializable, as they may be used as ordering methods in serializable data structures (like TreeSet, TreeMap). In order for the data structure to serialize successfully, the comparator (if provided) must implement Serializable. I agree on the class name to be more specific.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: This is wrapped in a synchronized block, so we should be ok.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: Done.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: Done

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
gskchaitanya: We initialize this array with the index number as the value for each entry, so we can't use Arrays.copyOf. We use this backup array to track the tagSets while iterating over a sorted cursor.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
gskchaitanya: Since this method is not implemented for the histograms yet, I turned off clover. Once we have agreed on this approach, I'm going to implement this.

lonemeow: Please add a comment explaining that under the `CLOVER:OFF` for now.

gskchaitanya: Done

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
lonemeow: Despite the synchronized block there's a race where one thread can get wrong value for `tagIndex` when another thread compacts the array. This used to rely heavily on the fact that `usedOffset` could never decrease, making it safe for that requires some thought.

gskchaitanya: We would keep incrementing the `usedOffset` and enter this block only when `usedOffset` value is more than `tagSets` length. One thread would get a chance to expand the array and the other thread waiting on the lock will check for the array length again and since the array is expanded, it will break out of the condition. Given we do not shrink the array, do you think of a case the race would still happen?

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -161,13 +161,18 @@ private Cursor newCursor(boolean sorted) {
     private int i = -1;
     private long base = 0;
     private int[] table;
+    private final Integer[] sortedRef;
```
manolama, line 164: If we're not doing null checks on the values, make it a primitive array to save some memory.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -161,13 +161,18 @@ private Cursor newCursor(boolean sorted) {
...
+      for (int i = 0; i < tagSets.length; i++) {
+        sortedRef[i] = i;
+      }
+      this.tagSets = tagSets;
       if (sorted) {
-        this.tagSets = tagSets.clone();
-        Arrays.sort(this.tagSets, TagSetsHelper::compare);
-      } else {
-        this.tagSets = tagSets;
+        TagSetsComparator comparator = new TagSetsComparator(tagSets);
```
manolama, line 174: Make the TagSetsComparator a static class that implements `compare()` and use a static instance of it. That way we don't pass the tag sets to the comparator to the ctor and can avoid instantiating one of these every time we run the cursor.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -576,19 +596,28 @@ long hashCode(final String[] tags) {
...
 
+    @SuppressFBWarnings(
+        value = {"EI_EXPOSE_REP2"},
+        justification = "Avoid creating copies for performance reasons.")
     private CursorImpl(final String[][] tagSets, final String[] fields, final Type[] types,
         final boolean sorted) {
       this.fields = fields;
       this.types = types;
+      // Holds the indexes of tagSets in the case of sorted iteration
+      sortedRef = new Integer[tagSets.length];
```
manolama, line 610: ditto here if nulls aren't needed. Save the object overhead.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -576,19 +596,28 @@ long hashCode(final String[] tags) {
...
+      for (int i = 0; i < tagSets.length; i++) {
+        sortedRef[i] = i;
+      }
+      this.tagSets = tagSets;
       if (sorted) {
-        this.tagSets = tagSets.clone();
-        Arrays.sort(this.tagSets, TagSetsHelper::compare);
-      } else {
-        this.tagSets = tagSets;
+        TagSetsComparator comparator = new TagSetsComparator(tagSets);
```
manolama, line 616: and static comparator here too. looks like it was `TagSetsHelper::compare`

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -597,22 +626,31 @@ public String getMetricId() {
...
+        }
+      }
 
-      long index = index(tagSets[i], true);
+      long index = index(currentTagSet, true);
 
       if (NOT_FOUND == index) {
         LOGGER.error("Missing index on Read. Tags: {}. Concurrency error or bug",
-            Arrays.asList(tagSets[i]));
+            Arrays.asList(currentTagSet));
```
manolama, line 646: Depending on how often this happens we should do `Arrays.toString()`

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -204,24 +209,30 @@ public boolean next() {
...
       }
-      return tagSets[i];
+      return tagSets[sortedRef[i]];
     }
-
+ ///CLOVER:OFF
+    // Will turn on Clover once this method is implemented.
+    @Override
+    public String[] freeCurrentRow() {
+      return null;
```
manolama, line 221: throw an `UnsupportedOperationException` just to save ourselves.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
```diff
@@ -294,11 +296,29 @@ long index(final String[] tags, final boolean isReading) {
...
                             : oldLength * 2;
-                    tagSets = Arrays.copyOf(tagSets, newLength);
+                    String[][] newTagSet = new String[newLength][];
+                    int cnt = 0;
+                    for (int i = 0; i < tagSets.length; i++) {
+                      // Skip the entries that would have been freed up (stale)
+                      // System.arraycopy would be faster, but can't use it here due to 
+                      // the conditional inclusion 
+                      String[] tagSeti = tagSets[i];
+                      if ((tagSeti.length > 0 && tagSeti[0] != null)
```
manolama, line 312: are tag sets guaranteed to be non-null here? Or do we need a null check for safety?

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
manolama: The DCLP is fine here and protects the assignment. But now won't the array just keep growing forever? Albeit with some empty slots? We also need a way to actually compact the array and relieve pressure.

The bigger issue GSK and I talked about is that the nulling of the tag set is done by the reporter and the tag set is shared amongst reporters so one could null it while the other doesn't and that would goof things up. So there is a bit of work to be redone here.

manolama (changes requested):

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: Not only memory but performance and creation of unnecessary objects. We need to minimize pressure on garbage collection

gskchaitanya: The previous implementation is creating a clone of the tagSets (`tagSets.clone();`) and sorts the tagsets so the iterator calling the cursor will iterate over the sorted list that's needed for the metric registry to group metrics of the same tagset. 
The current implementation avoids cloning the tagSets (and sorting it) and instead sorts just the Integer array that contains the indices of the tagSets (see https://github.com/ultrabrew/metrics/pull/28/files#diff-fa8eef85c6812afee9ae6b074bdc7cbfR47-R48). When the metric registry needs to iterate over the tagSets in sorted order, it simply goes over the sorted indices array (the `Integer[] `in this case). Now, I wanted to use the `Arrays.sort `functionality since it's the most efficient sort library that optimizes and runs various algorithms depending on the array length. 

The other reason to use Object array instead of the native array is that since I wanted to use the `Arrays.sort`, it needs `java.util.Comparator` and the Comparator can only work with Objects.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
gskchaitanya: This comparator doesn't actually sort the `tagSets` but sorts the indices of `tagSets` ( through `sortedRef`, refer the logic at https://github.com/ultrabrew/metrics/pull/28/files#diff-fa8eef85c6812afee9ae6b074bdc7cbfR47-R48). This is done to avoid cloning the whole `tagSets` which is several orders magnitude expensive than the `sortedRef` array. So, a static method doesn't work in this case since I'm using `Arrays.sort` and that uses a comparator.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
gskchaitanya: I intend to implement this once we agreed on the logic to purge the stale entries, so left it to null for now. Will change it to `UnsupportedOperationException` temporarily as you suggested :)

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: yes, they will be non-null.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: The `tagSets` array will be compacted by excluding the elements that are marked for deletion.
https://github.com/ultrabrew/metrics/pull/28/files#diff-fdded27fe21d54ab627b000d6b884cc8R320
and https://github.com/ultrabrew/metrics/pull/28/files#diff-fdded27fe21d54ab627b000d6b884cc8R320

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: Discussed in the previous comment

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
lonemeow: Consider this interleaving of events:

Preconditions:
 `tagSets.length == 2`
 `usedOffset == 2`
 `tagSets` contains 1 stale entry

Thread 1:

```java
int tagIndex = unsafe.getAndAddInt(this, usedOffset, 1);
// tagIndex == 2
```

Thread 2:

```java
int tagIndex = unsafe.getAndAddInt(this, usedOffset, 1);
// tagIndex == 3 
if (tagIndex >= tagSets.length) { 
  synchronized (this) { 
    if (tagIndex >= tagSets.length) { 
                  
      boolean isEqual = tagIndex == tagSets.length;
      final int oldLength = tagSets.length;
      final int newLength = 
          oldLength > TAGSETS_MAX_INCREMENT ? oldLength + TAGSETS_MAX_INCREMENT
              : oldLength * 2;
      String[][] newTagSet = new String[newLength][];
      int cnt = 0;
      for (int i = 0; i < tagSets.length; i++) { 
        String[] tagSeti = tagSets[i];
        if ((tagSeti.length > 0 && tagSeti[0] != null)
            || (tagSeti.length == 0)) { 
          String[] newTag = tagSeti;
          newTagSet[cnt++] = newTag;
        } 
      } 
      // cnt == 1, isEqual == false
      unsafe.putIntVolatile(this, usedOffset, isEqual ? cnt + 1 : cnt + 2);
      // usedOffset == 2 
      tagSets = newTagSet;
      tagIndex = isEqual ? cnt : cnt + 1;
      // tagIndex == 2 
    } 
  } 
} 
```


Thread 1:

```java
if (tagIndex >= tagSets.length) { 
  // skipped, condition is false
} 
```

Now both threads have `tagIndex == 2` and are about to write to `tagSets[tagIndex]`.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: What if tagsets changes while you hold the cursor to it? Previous implementation did not have this problem as it has a copy of it while iterating. Or has this been addressed already someway?

mmannerm: I would still prefer using primitive array instead of object array of objects. Lots of garbage. Cant search now but pretty sure there is some quicksort for primitive arrays also

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: Is it, because that is just creating one array of pointers to strings compared to creating an array of pointers and each pointer is again a new Integer object. I would not buy that argument without proof of performance

mmannerm: Array.copy is a shallow copy

mmannerm:
> The currently sorted cursor iterator works on a copy of the tag sets and this PR changes that to work on the same tag sets avoiding additional memory space.

I actually am not sure that you avoid additional memory space and actually introduce more objects and additional memory space than previous implementation. A JMH test that proves you actually do this would in order. The array.copy is shallow copy and points to same string arrays. Introducing an array of integer objects is the exact same memory (just the array) footprint , but every single Integer is an additional object compared to reusing the String[] objects already in memory. Or am I missing something?

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
gskchaitanya: @mmannerm you are right, a separate Integer array is not needed here. Will keep the original logic of `tagSets.clone();` since 2d array clone is a shallow copy.

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java
gskchaitanya: @lonemeow yes, there's a race as you mentioned. Since the array would potentially be shrinked, there's not a clean way to handle this scenario, will close this pr out for now and will open it up once I have a solution.

Issue #29, opened 2020-01-24
ConcurrentMonoid Tables copy tags by reference, not by value


arungupta:
The tags, when copied at the first occurrence, should be by value. Currently it happens by reference to the tags array provided by caller. This leads to mistakes when the caller changes the contents of the original tag array for object reuse purposes.

These lines seem to have the issue:
https://github.com/ultrabrew/metrics/blob/master/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidLongTable.java#L307
https://github.com/ultrabrew/metrics/blob/master/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java#L380

arungupta:
Proposed fix in https://github.com/ultrabrew/metrics/pull/31

arungupta:
Fixed in Release 0.5.0

PR #31, opened 2020-02-04
fix issue #29 by copying tags by value in ConcurrentMonoid tables

arungupta:
had to fix existing multicursor test after this due primitive arrays

### DIFF
--- a/core/src/test/java/io/ultrabrew/metrics/data/BasicGaugeAggregatorTest.java
+++ b/core/src/test/java/io/ultrabrew/metrics/data/BasicGaugeAggregatorTest.java
@@ -10,7 +10,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
-import java.lang.reflect.Array;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


core/src/test/java/io/ultrabrew/metrics/data/BasicGaugeAggregatorTest.java
```diff
@@ -10,6 +10,7 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.lang.reflect.Array;
```
smrutilal2: unused import

smrutilal2 (dismissed):

lonemeow (commented):
Have you measured the performance impact?

arungupta:
> Have you measured the performance impact?

No.. but we should as part of our benchmarking. The copy of tags only happens on first occurrence. The previous version is actually worse in theory because it encourages the caller to create new tag arrays every time, which is a problem especially when tags are derived from workloads at runtime.

This change allows the caller to reuse the same tag array and keep changing values, hopefully reducing garbage.

lonemeow (approved):

smrutilal2 (approved):

PR #32, opened 2020-02-07
avoid for each

smrutilal2:


arungupta (approved):

smrutilal2:
Java for each loop is a big source of garbage. Every time it records a value, it creates an instance of ArrayList Iterator. So, we should use the traditional index based for loop instead.

lonemeow (approved):

PR #37, opened 2020-12-08
Add validation around null tag strings and drop the measurements with a

manolama:
warning when such tags are present.
Update the reporter docs.

### DIFF
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
@@ -70,14 +70,14 @@ public class InfluxDBClient {
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
       if (Strings.isNullOrEmpty(tags[i])) {
-        LOGGER.warn("Null or empty tag key in tags array: " 
-            + Arrays.toString(tags) + " for measurement " + measurement);
+        LOGGER.warn("Null or empty tag key in tags array: {} for measurement {}", 
+          Arrays.toString(tags), measurement);
         byteBuffer.position(rollback);
         return;
       }
       if (Strings.isNullOrEmpty(tags[i + 1])) {
-        LOGGER.warn("Null or empty tag value in tags array: " 
-            + Arrays.toString(tags) + " for measurement " + measurement);
+        LOGGER.warn("Null or empty tag value in tags array: {} for measurement {}", 
+          Arrays.toString(tags), measurement);
         byteBuffer.position(rollback);
         return;
       }
@@ -94,14 +94,14 @@ public class InfluxDBClient {
         byteBuffer.put(COMMA);
       }
       if (Strings.isNullOrEmpty(fields[i])) {
-        LOGGER.warn("Null or empty field name in array: " 
-            + Arrays.toString(fields) + " for measurement " + measurement);
+        LOGGER.warn("Null or empty field name in array: {} for measurement {}", 
+          Arrays.toString(fields), measurement);
         byteBuffer.position(rollback);
         return;
       }
       if (Strings.isNullOrEmpty(fields[i + 1])) {
-        LOGGER.warn("Null or empty field value in array: " 
-            + Arrays.toString(fields) + " for measurement " + measurement);
+        LOGGER.warn("Null or empty field value in array: {} for measurement {}", 
+          Arrays.toString(fields), measurement);
         byteBuffer.position(rollback);
         return;
       }
@@ -167,4 +167,5 @@ public class InfluxDBClient {
       byteBuffer.clear();
     }
   }
+  
 }
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
@@ -4,6 +4,7 @@
 
 package io.ultrabrew.metrics.reporters.influxdb;
 
+import java.util.Arrays;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.BufferOverflowException;
@@ -17,12 +18,15 @@ import org.apache.http.entity.ContentType;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.http.util.EntityUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This class provides methods to access InfluxDB.
  */
 public class InfluxDBClient {
-
+  private static final Logger LOGGER = LoggerFactory.getLogger(InfluxDBClient.class);
+  
   private static final String UTF_8 = StandardCharsets.UTF_8.name();
   private static final byte WHITESPACE = ' ';
   private static final byte COMMA = ',';
@@ -57,8 +61,25 @@ public class InfluxDBClient {
   private void doWrite(final String measurement, final String[] tags, final String[] fields,
       final long timestamp)
       throws IOException {
+    if (measurement == null || measurement.isEmpty()) {
+      LOGGER.warn("Null or empty measurement.");
+      return;
+    }
+    int rollback = byteBuffer.position();
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
+      if (tags[i] == null || tags[i].isEmpty()) {
+        LOGGER.warn("Null or empty tag key in tags array: " 
+            + Arrays.toString(tags) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
+      }
+      if (tags[i + 1] == null || tags[i + 1].isEmpty()) {
+        LOGGER.warn("Null or empty tag value in tags array: " 
+            + Arrays.toString(tags) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
+      }
       byteBuffer.put(COMMA)
           .put(tags[i].getBytes(UTF_8))
           .put(EQUALS)
@@ -70,6 +91,18 @@ public class InfluxDBClient {
     for (int i = 0; i < fields.length; i += 2) {
       if (!f) {
         byteBuffer.put(COMMA);
+      }
+      if (fields[i] == null || fields[i].isEmpty()) {
+        LOGGER.warn("Null or empty field name in array: " 
+            + Arrays.toString(fields) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
+      }
+      if (fields[i + 1] == null || fields[i + 1].isEmpty()) {
+        LOGGER.warn("Null or empty field value in array: " 
+            + Arrays.toString(fields) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
       }
       byteBuffer.put(fields[i].getBytes(UTF_8))
           .put(EQUALS)
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
@@ -33,6 +33,7 @@ public class InfluxDBClient {
   private static final byte COMMA = ',';
   private static final byte EQUALS = '=';
   private static final byte NEWLINE = '\n';
+  private static final byte[] NULL_STRING = new byte[] { 'N', 'U', 'L', 'L' };
 
   private final ByteBuffer byteBuffer;
   private final URI dbUri;
@@ -76,15 +77,15 @@ public class InfluxDBClient {
         return;
       }
       if (Strings.isNullOrEmpty(tags[i + 1])) {
-        LOGGER.warn("Null or empty tag value in tags array: {} for measurement {}", 
-          Arrays.toString(tags), measurement);
-        byteBuffer.position(rollback);
-        return;
+        // TODO - Some users want this, some don't. Set a flag in the builder.
+        //LOGGER.warn("Null or empty tag value in tags array: {} for measurement {}", 
+        //  Arrays.toString(tags), measurement);
       }
       byteBuffer.put(COMMA)
           .put(tags[i].getBytes(UTF_8))
           .put(EQUALS)
-          .put(tags[i + 1].getBytes(UTF_8));
+          .put(Strings.isNullOrEmpty(tags[i + 1]) ? NULL_STRING : 
+            tags[i + 1].getBytes(UTF_8));
     }
     byteBuffer.put(WHITESPACE);
 
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
@@ -4,6 +4,7 @@
 
 package io.ultrabrew.metrics.reporters.influxdb;
 
+import io.ultrabrew.metrics.util.Strings;
 import java.util.Arrays;
 import java.io.IOException;
 import java.net.URI;
@@ -61,20 +62,20 @@ public class InfluxDBClient {
   private void doWrite(final String measurement, final String[] tags, final String[] fields,
       final long timestamp)
       throws IOException {
-    if (measurement == null || measurement.isEmpty()) {
+    if (Strings.isNullOrEmpty(measurement)) {
       LOGGER.warn("Null or empty measurement.");
       return;
     }
     int rollback = byteBuffer.position();
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
-      if (tags[i] == null || tags[i].isEmpty()) {
+      if (Strings.isNullOrEmpty(tags[i])) {
         LOGGER.warn("Null or empty tag key in tags array: " 
             + Arrays.toString(tags) + " for measurement " + measurement);
         byteBuffer.position(rollback);
         return;
       }
-      if (tags[i + 1] == null || tags[i + 1].isEmpty()) {
+      if (Strings.isNullOrEmpty(tags[i + 1])) {
         LOGGER.warn("Null or empty tag value in tags array: " 
             + Arrays.toString(tags) + " for measurement " + measurement);
         byteBuffer.position(rollback);
@@ -92,13 +93,13 @@ public class InfluxDBClient {
       if (!f) {
         byteBuffer.put(COMMA);
       }
-      if (fields[i] == null || fields[i].isEmpty()) {
+      if (Strings.isNullOrEmpty(fields[i])) {
         LOGGER.warn("Null or empty field name in array: " 
             + Arrays.toString(fields) + " for measurement " + measurement);
         byteBuffer.position(rollback);
         return;
       }
-      if (fields[i + 1] == null || fields[i + 1].isEmpty()) {
+      if (Strings.isNullOrEmpty(fields[i + 1])) {
         LOGGER.warn("Null or empty field value in array: " 
             + Arrays.toString(fields) + " for measurement " + measurement);
         byteBuffer.position(rollback);
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
@@ -10,6 +10,7 @@ import io.ultrabrew.metrics.data.Cursor;
 import io.ultrabrew.metrics.data.CursorEntry;
 import io.ultrabrew.metrics.data.Type;
 import io.ultrabrew.metrics.reporters.TimeWindowReporter;
+import io.ultrabrew.metrics.util.Strings;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Map;
@@ -150,11 +151,11 @@ public class InfluxDBReporter extends TimeWindowReporter {
       if (baseUri == null) {
         throw new IllegalArgumentException("Invalid baseUri");
       }
-      if ((database == null || database.isEmpty()) && 
-          (endpoint == null || endpoint.isEmpty())) {
+      if (Strings.isNullOrEmpty(database) && 
+          Strings.isNullOrEmpty(endpoint)) {
         throw new IllegalArgumentException("Invalid database");
       }
-      if (endpoint != null && !endpoint.isEmpty()) {
+      if (!Strings.isNullOrEmpty(endpoint)) {
         return new InfluxDBReporter(baseUri.resolve(endpoint), windowSeconds,
             bufferSize, defaultAggregators, metricAggregators);
       }
--- a/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
+++ b/reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
@@ -82,6 +82,7 @@ public class InfluxDBReporter extends TimeWindowReporter {
   public static class Builder extends TimeWindowReporterBuilder<Builder, InfluxDBReporter> {
 
     private URI baseUri = null;
+    private String endpoint = null;
     private String database = null;
     private int windowSeconds = 1;
     private int bufferSize = 64 * 1024;
@@ -96,6 +97,18 @@ public class InfluxDBReporter extends TimeWindowReporter {
      */
     public Builder withBaseUri(final URI baseUri) {
       this.baseUri = baseUri;
+      return this;
+    }
+    
+    /**
+     * An optional endpoint to build the URI that, when provided, overrides the 
+     * default `/write?db=$database`. You must supply the full URI string, e.g.
+     * `/api/put/influx/write?db=mydatabase`. 
+     * @param endpoint An optional non-null endpoint to apply to the base URI.
+     * @return The builder.
+     */
+    public Builder withEndpoint(final String endpoint) {
+      this.endpoint = endpoint;
       return this;
     }
 
@@ -137,8 +150,13 @@ public class InfluxDBReporter extends TimeWindowReporter {
       if (baseUri == null) {
         throw new IllegalArgumentException("Invalid baseUri");
       }
-      if (database == null || database.isEmpty()) {
+      if ((database == null || database.isEmpty()) && 
+          (endpoint == null || endpoint.isEmpty())) {
         throw new IllegalArgumentException("Invalid database");
+      }
+      if (endpoint != null && !endpoint.isEmpty()) {
+        return new InfluxDBReporter(baseUri.resolve(endpoint), windowSeconds,
+            bufferSize, defaultAggregators, metricAggregators);
       }
       return new InfluxDBReporter(baseUri.resolve("/write?db=" + database), windowSeconds,
           bufferSize, defaultAggregators, metricAggregators);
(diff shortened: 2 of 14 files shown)


reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -73,10 +75,33 @@ private CloseableHttpClient getHttpClient() {
         .build();
   }
 
+  @SuppressFBWarnings(value="NP", justification="Null check IS performed on tags")
   void write(final String metricName,
       final String[] tags,
       final long timestamp,
       final String value) throws IOException {
+    // validation
+    if (metricName == null || metricName.isEmpty()) {
```
lonemeow: This "null or empty" is repeated so many times that it would probably make sense to add a helper method for it.

manolama: If we have Guava already we can use the `Strings.isNullorEmpty` helper. Or maybe some Apache lib has it. I'll poke around. Don't want to add a new dependency.

smrutilal2: Let's put this as a utility method.
```
 public static boolean isNullOrEmpty(String s) {
    return s == null || s.isEmpty();
  }
```

mmannerm: Absolutely no Guava anywhere near this library

smrutilal2 (dismissed):

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
manolama: @lonemeow Fixed, mind reviewing please?

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
```diff
@@ -137,9 +150,14 @@ public InfluxDBReporter build() {
       if (baseUri == null) {
         throw new IllegalArgumentException("Invalid baseUri");
       }
-      if (database == null || database.isEmpty()) {
+      if ((database == null || database.isEmpty()) && 
+          (endpoint == null || endpoint.isEmpty())) {
```
lonemeow: You could use the new `isNullOrEmpty()` helper here.

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
```diff
@@ -137,9 +150,14 @@ public InfluxDBReporter build() {
...
         throw new IllegalArgumentException("Invalid database");
       }
+      if (endpoint != null && !endpoint.isEmpty()) {
```
lonemeow: You could use the new `isNullOrEmpty()` helper here.

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
```diff
@@ -86,6 +87,23 @@ public void testWriteFailsManualFlush() throws Exception {
...
+    }};
+    OpenTSDBHttpClient client = new OpenTSDBHttpClient(DUMMY_DB_URI, 64, true);
+    client.write(null, new String[] { "host", "web01" }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { null, "web01" }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host", null }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host", "web01" }, 1534055562000000003L, null);
+    client.write("cpu_load_short.temp", null, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host" }, 1534055562000000003L, "80");
+    new Verifications() {{
+      times = 0;
```
lonemeow: Is there something missing here? What are we verifying was called zero times?

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
```diff
@@ -86,6 +87,23 @@ public void testWriteFailsManualFlush() throws Exception {
     assertEquals(0, client.currentBatchSize);
   }
 
+  @Test
+  public void testNullStringsAndTagsValidation() throws Exception {
+    new Expectations() {{
+
```
lonemeow: If there are no expectations, just remove the entire block.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporterTest.java
```diff
@@ -64,6 +64,20 @@ public void testSetBufferSize() {
...
+  public void testSeEndpoint() {
+    InfluxDBReporter r = InfluxDBReporter.builder()
+        .withBaseUri(TEST_URI)
+        .withDatabase("test") // ignored
+        .withEndpoint("/my/change?db=foo")
+        .withBufferSize(12765)
+        .build();
+
+    InfluxDBClient c = Deencapsulation.getField(r, "dbClient");
+    URI uri = Deencapsulation.getField(c, "dbUri");
```
lonemeow: Please don't add dependency on `Deencapsulation` since it's deprecated and removed from newer versions of `jmockit`.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
```diff
@@ -147,6 +147,22 @@ public void testWriteSplitting() throws IOException {
...
+  @Test
+  public void testNullStrings() throws Exception {
+    new Expectations() {{
+      
+    }};
+    client.write(null, new String[] { "host", "web01" }, new String[] { "temp", "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { null, "web01" }, new String[] { "temp", "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { "host", null }, new String[] { "temp", "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { "host", "web01" }, new String[] { null, "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { "host", "web01" }, new String[] { "temp", null }, 1534055562000000003L);
```
lonemeow, line 169: You should probably test a case where not all the data is invalid to verify the buffer position rollback works correctly.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
```diff
@@ -147,6 +147,22 @@ public void testWriteSplitting() throws IOException {
     }};
   }
 
+  @Test
+  public void testNullStrings() throws Exception {
+    new Expectations() {{
+
```
lonemeow: If there are no expectations, just remove the entire block.

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -57,8 +62,25 @@ private CloseableHttpClient getHttpClient() {
...
+      LOGGER.warn("Null or empty measurement.");
+      return;
+    }
+    int rollback = byteBuffer.position();
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
+      if (Strings.isNullOrEmpty(tags[i])) {
+        LOGGER.warn("Null or empty tag key in tags array: " 
+            + Arrays.toString(tags) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
```
lonemeow, line 76: Instead of adding the rollback to each check, it could be done with `try`-`finally` to make the code maybe a little less error prone, though I didn't actually try writing it to see if it would look better that way.

manolama: I think that'd be a bit uglier as I don't want to rollback if we have valid data. Or I'd have to get the position and reset it to the same position in the finally.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporterTest.java
manolama: I was just using what's already there in the other tests. Is there a different way to get to the private state?

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
lonemeow: Or use `-1` or similar impossible position as marker for no rollback, yes. I'm just slightly concerned that this way it's entirely possible to forget the rollback call later when adding more checks. In the end it's a tradeoff, choosing which way produces the more readable and robust result.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporterTest.java
lonemeow: No, and there should not be; verifying private state is not good testing practice, tests should verify externally visible behavior.

I know some of the old tests do funky things, and plan to fix those, so I'd prefer if you don't add more such cases.

lonemeow: In this case, what we want to test is that setting the endpoint actually changes the URI being called, so you'd want to mock the HTTP client and verify that it is called to perform a request to the correct URI.

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -57,8 +62,25 @@ private CloseableHttpClient getHttpClient() {
...
       throws IOException {
+    if (Strings.isNullOrEmpty(measurement)) {
+      LOGGER.warn("Null or empty measurement.");
+      return;
+    }
+    int rollback = byteBuffer.position();
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
+      if (Strings.isNullOrEmpty(tags[i])) {
+        LOGGER.warn("Null or empty tag key in tags array: "
```
mmannerm:
```suggestion
        LOGGER.warn("Null or empty tag key in tags array: {} for measurement {}", Arrays.toString(tags), measurement); 
```

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -57,8 +62,25 @@ private CloseableHttpClient getHttpClient() {
...
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
+      if (Strings.isNullOrEmpty(tags[i])) {
+        LOGGER.warn("Null or empty tag key in tags array: " 
+            + Arrays.toString(tags) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
+      }
+      if (Strings.isNullOrEmpty(tags[i + 1])) {
+        LOGGER.warn("Null or empty tag value in tags array: "
```
mmannerm:
```suggestion
        LOGGER.warn("Null or empty tag value in tags array: {} for measurement {}", Arrays.toString(tags), measurement); 
```

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -71,6 +93,18 @@ private void doWrite(final String measurement, final String[] tags, final String
       if (!f) {
         byteBuffer.put(COMMA);
       }
+      if (Strings.isNullOrEmpty(fields[i])) {
+        LOGGER.warn("Null or empty field name in array: "
```
mmannerm:
```suggestion
        LOGGER.warn("Null or empty field name in array: {} for measurement {}", Arrays.toString(fields), measurement); 
```

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -71,6 +93,18 @@ private void doWrite(final String measurement, final String[] tags, final String
...
         byteBuffer.put(COMMA);
       }
+      if (Strings.isNullOrEmpty(fields[i])) {
+        LOGGER.warn("Null or empty field name in array: " 
+            + Arrays.toString(fields) + " for measurement " + measurement);
+        byteBuffer.position(rollback);
+        return;
+      }
+      if (Strings.isNullOrEmpty(fields[i + 1])) {
+        LOGGER.warn("Null or empty field value in array: "
```
mmannerm:
```suggestion
        LOGGER.warn("Null or empty field value in array: {} for measurement {}", Arrays.toString(fields), measurement); 
```

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -77,6 +80,30 @@ void write(final String metricName,
...
+    if (Strings.isNullOrEmpty(value)) {
+      LOG.warn("Null or empty value.");
+      return;
+    }
+    if (tags == null) {
+      LOG.warn("At least one tag pair must be present.");
+      return;
+    }
+    if (tags.length % 2 != 0) {
+      LOG.warn("Uneven tag count: " + Arrays.toString(tags) + " for metric " + metricName);
```
mmannerm:
```suggestion
      LOG.warn("Uneven tag count: {} for metric {}", Arrays.toString(tags), metricName);
```

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -77,6 +80,30 @@ void write(final String metricName,
...
+      return;
+    }
+    for (int i = 0; i < tags.length; i++) {
+      if (Strings.isNullOrEmpty(tags[i])) {
+        LOG.warn("Null tag key or value in: " + Arrays.toString(tags) + " for metric " + metricName);
```
mmannerm:
```suggestion
        LOG.warn("Null tag key or value in: {} for metric {}", Arrays.toString(tags), metricName);
```

mmannerm:
In general, what is the reasoning why tag values of null are not allowed. I understand why tag keys as null should not be allowed, but not so sure about tag values not null. I would argue it is better for metrics library API to allow such string arrays with null tag values, as it allows easy and fast generation of such tag arrays without conditional building.

mmannerm:
Even if you would filter out the tag arrays with tag value as NULL, you should report the metric with all the other tags, not just drop the full measurement on the ground. I don’t understand this PR.

mmannerm:
or convert null to ”null” in reporters that don’t understand nulls.

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
manolama: yup missing, doh.

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
manolama: done.

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporter.java
manolama: Done.

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
manolama: done

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
manolama: done

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
manolama: good point.

manolama:
The impetus for this was that when I was testing some code I found the reporter was throwing these NPEs and I didn't know what was going on. Most time series stores do not allow null tag values. Some do allow for a stand-alone label but that is a separate data set than a key/value tags set. I had an array value that wasn't set on accident so now I'll be able to figure out where that's coming from and fix it.

Dropping the tag key isn't a good option as we don't know the user's intent. Maybe they intended to have a value there or they wanted to write "null" on purpose.

For our in-house stuff where we started seeing missing values and users were curious, we did decide to substitute a string for missing values depending on the library. We can put in "null" here if you want or make it an option in the reporter builder.

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
manolama: Fixed all the logs.

mmannerm:
I would personally convert it in the reporter to string value ”null” or similar to keep the interface clean and fast on reporting metrics. I.e. I would not have to in my filter or code that sends the metrics filter out variables that are null from string array and can just do

```
 metric.inc([ TAG_KEY1, tagValue1, TAG_KEY2, tagValue2 ]);
```

Instead of using dozens of lines of code to grow array, re-allocate objects in case the tagValue1 or tagValue2 or both are null.

manolama:
Updated @mmannerm

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
```diff
@@ -57,12 +63,29 @@ private CloseableHttpClient getHttpClient() {
...
       throws IOException {
+    if (Strings.isNullOrEmpty(measurement)) {
+      LOGGER.warn("Null or empty measurement.");
+      return;
+    }
+    int rollback = byteBuffer.position();
     byteBuffer.put(measurement.getBytes(UTF_8));
     for (int i = 0; i < tags.length; i += 2) {
+      if (Strings.isNullOrEmpty(tags[i])) {
+        LOGGER.warn("Null or empty tag key in tags array: {} for measurement {}",
```
lonemeow, line 74: This has potential to cause a _lot_ of log messages. I'm not sure if we should do some rate limiting or something here (applies to all the logging here of course).

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
```diff
@@ -147,6 +147,33 @@ public void testWriteSplitting() throws IOException {
...
+    client.write("cpu_load_short", new String[] { null, "web01" }, new String[] { "temp", "80" }, 1534055562000000003L);
+    // one good one in the middle
+    client.write("cpu_load_short", new String[] { "host", "web01" }, new String[] { "temp", "80" }, 1534055562000000003L);
+    // this is ok too
+    client.write("cpu_load_short", new String[] { "host", null }, new String[] { "temp", "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { "host", "web01" }, new String[] { null, "80" }, 1534055562000000003L);
+    client.write("cpu_load_short", new String[] { "host", "web01" }, new String[] { "temp", null }, 1534055562000000003L);
+    client.flush();
+    new Verifications() {{
+      httpClient.execute((HttpUriRequest) any);
```
lonemeow: Assert something about the payload sent here, to verify that it contains the expected data?

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -4,6 +4,9 @@
 
 package io.ultrabrew.metrics.reporters.opentsdb;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
```
lonemeow: Unused import?

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -77,6 +80,30 @@ void write(final String metricName,
       final String[] tags,
       final long timestamp,
       final String value) throws IOException {
+    // validation
+    if (Strings.isNullOrEmpty(metricName)) {
+      LOG.warn("Null or empty metric name.");
```
lonemeow, line 84: Same comment as the other similar place, these logs have potential to generate a _lot_ of messages, I wonder if we should have some kind of rate limiting in place?

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
```diff
@@ -86,6 +87,17 @@ public void testWriteFailsManualFlush() throws Exception {
...
 
+  @Test
+  public void testNullStringsAndTagsValidation() throws Exception {
+    OpenTSDBHttpClient client = new OpenTSDBHttpClient(DUMMY_DB_URI, 64, true);
+    client.write(null, new String[] { "host", "web01" }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { null, "web01" }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host", null }, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host", "web01" }, 1534055562000000003L, null);
+    client.write("cpu_load_short.temp", null, 1534055562000000003L, "80");
+    client.write("cpu_load_short.temp", new String[] { "host" }, 1534055562000000003L, "80");
```
lonemeow, line 109: Shouldn't this test verify something about the resulting behavior?

reporter-influxdb/src/main/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClient.java
manolama: This should rarely happen that we have a null tag key or field or value. If we do then it is a user bug so I'm ok flooding their logs with such messages. The one about null tag values would have been potentially much more naggy but since Mika wanted to substitute the string "null" it isn't an issue.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
manolama: yeah I just have to dig into Mockit and figure out how to parse the payload.

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
manolama: yup.

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
manolama: comment up there.

reporter-opentsdb/src/test/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClientTest.java
manolama: oh darn lost that bit, will restore it.

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBClientTest.java
lonemeow: Something along these lines should work:
```java
    new Verifications() {{
      HttpPost request;
      httpClient.execute(request = withCapture());
      assertThat(request.getEntity().someProperty(), is(someExpectedValue));
}};
```

reporter-influxdb/src/test/java/io/ultrabrew/metrics/reporters/influxdb/InfluxDBReporterTest.java
manolama: done for this test. The others still need protected getters for their private values in the client.

reporter-opentsdb/src/main/java/io/ultrabrew/metrics/reporters/opentsdb/OpenTSDBHttpClient.java
```diff
@@ -77,6 +79,30 @@ void write(final String metricName,
...
+    if (tags == null) {
+      LOG.warn("At least one tag pair must be present.");
+      return;
+    }
+    if (tags.length % 2 != 0) {
+      LOG.warn("Uneven tag count: {} for metric {}", Arrays.toString(tags), metricName);
+      return;
+    }
+    for (int i = 0; i < tags.length; i++) {
+      if (Strings.isNullOrEmpty(tags[i])) {
```
lonemeow: In `InfluxDBReporter` `null` tag _values_ are sent as a fixed `NULL` string, should that be the case here also?

manolama: No, null keys should always log an error as it's an invalid situation.

lonemeow: But won't this throw for both keys _and_ values?

manolama: ah whoops sorry this is the TSD reporter. yeah I'll add the null for values too I 'spose.

lonemeow (approved):

smrutilal2 (approved):

PR #39, opened 2021-02-19
log metric id and capacity

smrutilal2:

### DIFF
--- a/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
@@ -43,7 +43,6 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
   private static final String[] AGGREGATION_FIELDS = {"count", "sum", "min", "max", "lastValue"};
   private static final long[] IDENTITY = {0L, 0L, Long.MAX_VALUE, Long.MIN_VALUE, 0L};
 
-  private final String metricId;
   private final DistributionBucket buckets;
 
   private final String[] fields;
@@ -82,8 +81,7 @@ public class BasicHistogramAggregator extends ConcurrentMonoidIntTable implement
    */
   public BasicHistogramAggregator(final String metricId, final DistributionBucket buckets,
       final int maxCardinality, final int cardinality) {
-    super(AGGREGATION_FIELDS.length + buckets.getCount(), maxCardinality, cardinality, IDENTITY);
-    this.metricId = metricId;
+    super(metricId, AGGREGATION_FIELDS.length + buckets.getCount(), maxCardinality, cardinality, IDENTITY);
     this.buckets = buckets;
     this.fields = buildFields();
     this.types = buildTypes();
--- a/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
@@ -4,15 +4,14 @@
 
 package io.ultrabrew.metrics.data;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import sun.misc.Unsafe;
-
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
 
 public abstract class ConcurrentMonoidIntTable {
 
--- a/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
+++ b/core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
@@ -4,14 +4,15 @@
 
 package io.ultrabrew.metrics.data;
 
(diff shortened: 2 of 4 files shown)


arungupta (dismissed):

arungupta:
@smrutilal2 -- we should rate limit the logging as discussed earlier

core/src/main/java/io/ultrabrew/metrics/data/ConcurrentMonoidIntTable.java
```diff
@@ -4,14 +4,15 @@
 
 package io.ultrabrew.metrics.data;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
```
lonemeow: Can you undo the non-semantic style changes. If we want to change the style, that should be a separate discussion - let's try to keep the PRs to the actual changes for now.

lonemeow (dismissed):

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
```diff
@@ -82,8 +81,7 @@ public BasicHistogramAggregator(final String metricId, final DistributionBucket
    */
   public BasicHistogramAggregator(final String metricId, final DistributionBucket buckets,
```
arungupta, line 82: `metricId` should probably be renamed `metricName` since it's not an internal id but the actual name from the user.

arungupta (dismissed):

core/src/main/java/io/ultrabrew/metrics/data/CursorEntry.java
```diff
@@ -11,7 +11,7 @@
    *
    * @return identifier of the metric
    */
-  String getMetricId();
+  String getMetricName();
```
lonemeow: This changes ABI, breaks all existing users that have code dealing with this class, do we _really_ want to go there with this change?

smrutilal2: That's right. The change to `CursorEntry` would break the reporters. May be I will revert this class.

arungupta: hmm.. yeah i was about to mention that as well. we cannot change anything that's part of public api. @smrutilal2 i guess let's revert it for now.

smrutilal2: @arungupta and @lonemeow, I have reverted `CursorEntry`. Please take a look.

core/src/main/java/io/ultrabrew/metrics/data/BasicCounterAggregator.java
```diff
@@ -37,35 +37,35 @@ public BasicCounterAggregator(final Counter counter) {
   /**
    * Create a monoid for common aggregation functions for a Counter.
    *
-   * @param metricId identifier of the metric associated with this aggregator
+   * @param metricName identifier of the metric associated with this aggregator
```
lonemeow: Should the Javadoc also be updated to say "name" instead of "identifier" if we're going for a terminology change?

smrutilal2: Done.

core/src/main/java/io/ultrabrew/metrics/data/BasicHistogramAggregator.java
mmannerm: Do not change the public interface with backwards incompatible changes. Also identifier was chosen instead of name for a reason as it is an identifier, not just a label

core/src/main/java/io/ultrabrew/metrics/data/BasicCounterAggregator.java
mmannerm: Why do you want to change from a specific term describing the intent and purpose of the variable to a name that is NOT precise and can be misunderstood as a label type instead of what it truly is: an identifier. Also this breaks consistency and backwards compatibility in the library. Do not make this change

mmannerm:
Please do not make backwards incompatible changes and use consistent terminology. Metric identifiers are called identifiers in all documentation, interfaces etc.

mmannerm:
https://github.com/ultrabrew/metrics/blob/c251e99aa71a431d914abf394db95b1b8a058c9b/core/src/main/java/io/ultrabrew/metrics/Metric.java#L43

mmannerm (changes requested):
Please revert naming changes

core/src/main/java/io/ultrabrew/metrics/data/BasicCounterAggregator.java
smrutilal2: It's not an internal id but the actual name of the metric provided by the user. So, I was trying to be consistent with the semantics. But, I agree that it will break consistency. I will revert my last commit.

smrutilal2:
> Please revert naming changes

Reverting it. Thank you!

smrutilal2:
Hi have reverted the last commit. Please take a look.
@mmannerm, @lonemeow and @arungupta .

arungupta (approved):

core/src/main/java/io/ultrabrew/metrics/data/BasicCounterAggregator.java
mmannerm: There is no such things as internal id, there is public id of the metric that the developer defines.

mmannerm (approved):

PR #41, opened 2021-04-22
publish to sonatype repo

### LINKED ISSUE #14: Improve generated pom.xml
Currently the `pom.xml` generated is missing most optional fields such as license, this should be improved.

smrutilal2:


examples/webapp/build.gradle
```diff
@@ -7,11 +7,11 @@ publishMavenJavaPublicationToMavenRepository {
 }
 
 dependencies {
-    compile project(':core')
+    api project(':core')
```
lonemeow: Technically, this is `implementation` I'd think.

build.gradle
```diff
@@ -121,32 +111,75 @@ subprojects {
...
         repositories {
             maven {
-                def artifactory = 'https://oss.jfrog.org/artifactory/oss-snapshot-local'
-                def bintray = 'https://api.bintray.com/maven/ultrabrew/m2/' + rootProject.name + '/;publish=1'
+                def snapshotRepository = 'https://s01.oss.sonatype.org/content/repositories/snapshots/'
+                def stagingRepository = 'https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/'
 
                 afterEvaluate {
-                    url = version.toString().endsWith('-SNAPSHOT') ? artifactory : bintray
+                    url = version.toString().endsWith('-SNAPSHOT') ? snapshotRepository : stagingRepository
```
lonemeow, line 164: How does the artifact end up in public repo from the staging repo?

core/build.gradle
```diff
@@ -1,8 +1,11 @@
+description = 'A lightweight, high-performance Java instrumentation library'
+
 dependencies {
-    compile group: 'org.slf4j', name: 'slf4j-api', version:'1.7.25'
+    api group: 'org.slf4j', name: 'slf4j-api', version:'1.7.25'
```
lonemeow, line 4: Hmm, is there need for this to be `api`? I don't remember any slf4j classes being visible in the public API but I might remember wrong...

build.gradle
```diff
@@ -83,34 +84,23 @@ subprojects {
...
         format = 'ALL'
         failBuildOnCVSS = 4.0
-    }
-
-    spotbugs {
-        sourceSets = [sourceSets.main]
+        cve {
+            urlModified = "https://freedumbytes.gitlab.io/setup/nist-nvd-mirror/nvdcve-1.1-modified.json.gz"
+            urlBase = "https://freedumbytes.gitlab.io/setup/nist-nvd-mirror/nvdcve-1.1-%d.json.gz"
+        }
```
lonemeow: This is to work around the broken NVD files? Worth a comment to note that.

build.gradle
smrutilal2: It's manually done from the staging repo. It also can be automated so that every build is published to the central repo. But, I kept it manual to start with. Once we verify that it's building fine, we can automate the promotion.

core/build.gradle
smrutilal2: It's needed because of the SLF4JReporter. Otherwise it will give compilation error to the clients unless they add this dependency explicitly.

build.gradle
smrutilal2: Yes, it's broken since yesterday. I will add a comment.

examples/webapp/build.gradle
smrutilal2: sure.

build.gradle
smrutilal2: Ref: https://central.sonatype.org/publish/release/

lonemeow: Ideally we'd automatically promote releases that are tagged as releases in git, I suppose.

core/build.gradle
lonemeow: Oh, that's in core. Then it is indeed `api` scoped.

build.gradle
smrutilal2: Yes, I will add that in my next pr may be.

lonemeow (approved):

arungupta (approved):

PR #47, opened 2021-05-18
Add support for Distribution bucket for a double valued metric

### LINKED ISSUE #27: Type safe Histograms
Currently, the Histograms stores only Integer values. This can result in bad values, if it's attached to a `GaugeDouble` metric. Because, user will store a `double` value and the reporter will report the long bits of the double value which is garbage. So, we need to added support Doubles for Histograms and use it appropriately.

### LINKED ISSUE #46: Add support for double valued distribution buckets.
Currently we only support the `long` values for the distribution buckets. We need to support `double` values as well. 
User can configure the double valued buckets like `[0.0, 0.1, 1.5, 50.0]`.

smrutilal2:

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,9 +289,9 @@ There two types of distribution buckets available:
 - `DoubleValuedDistributionBucket` represented by a primitive `double` array
   
 ##### DistributionBucket
-Used to represent the distribution of a integer value. For example time spent in milliseconds or size of a messaging queue.
+Used to represent the distribution of an integer value. For example time spent in milliseconds or size of a messaging queue.
 
-For a given distribution array: [0, 10, 100, 500, 1000], the buckets would be like:
+For a given array of latency distribution in milliseconds [0, 10, 100, 500, 1000], the buckets would be like:
 * [0,10) for values 0-9
 * [10,100) for values 10-99
 * [100,500) for values 100-499
@@ -312,19 +312,20 @@ For a given distribution array: [0, 10, 100, 500, 1000], the buckets would be li
 
   Timer timer = metricRegistry.timer(metricId);    // creates a timer metric with id "latency"
 
-  long start = Timer.start();
+  long start = System.currentTimeMillis();
   // doSomething();
-  timer.stop(start, tagset); // records the latency and the distribution.
+  long end = System.currentTimeMillis();  
+  timer.update(end - start, tagset); // records the latency and the distribution in millisecond.
 ```
 
 ##### DoubleValuedDistributionBucket
-Used to represent the distribution of a floating point value. For example ads auction price.
+Used to represent the distribution of a double-precision floating point value. For example ads auction price.
 
 For a given distribution array: [0.0, 0.25, 0.5, 1.0, 10.0], the buckets would be like:
-* [0.0, 0.25) for values between 0.0-0.24
-* [0.25, 0.5) for values 0.25-0.4
-* [0.5, 1.0) for values 0.5-0.9
-* [1.0, 10.0) for values 1.0-9.9
+* [0.0, 0.25) for 0.0 <= value < 0.25
+* [0.25, 0.5) for 0.25 <= value < 0.5
+* [0.5, 1.0) for 0.5 <= value < 1.0
+* [1.0, 10.0) for 1.0 <= value < 10.0
 * overflow  for values  >= 10.0
 * underflow for values  < 0.0
 
@@ -339,7 +340,7 @@ For a given distribution array: [0.0, 0.25, 0.5, 1.0, 10.0], the buckets would b
 
   String[] tagset = new String[] {"experiment", "exp1"};
 
-  GuageDouble autionPrice = metricRegistry.gaugeDouble(metricId);    // creates a guage double metric with id "auction_price"
+  GaugeDouble autionPrice = metricRegistry.gaugeDouble(metricId);    // creates a guage double metric with id "auction_price"
 
   double price = getAuctionPrice();
   autionPrice.set(price, tagset1); // records the auction_price and the distribution.
--- a/README.md
+++ b/README.md
@@ -291,17 +291,17 @@ There two types of distribution buckets available:
 ##### DistributionBucket
 Used to represent the distribution of an integer value. For example time spent in milliseconds or size of a messaging queue.
 
-For a given array of latency distribution in milliseconds [0, 10, 100, 500, 1000], the buckets would be like:
-* [0,10) for values 0-9
-* [10,100) for values 10-99
-* [100,500) for values 100-499
-* [500,1000) for values 500-999
-* overflow  for values  >= 1000
+For a given array of latency distribution in nanoseconds [0, 10_000_000, 100_000_000, 500_000_000, 1000_000_000], the buckets would be like:
+* [0,10_000_000) for values 0 - 9_999_999
+* [10_000_000,100_000_000) for values 10_000_000 - 99_999_999
+* [100_000_000,500_000_000) for values 100_000_000 - 499_999_999
+* [500_000_000,1000_000_000) for values 500_000_000 - 999999999
+* overflow  for values  >= 1000_000_000
 * underflow for values  < 0
 
 ```Java
   String metricId = "latency";
-  DistributionBucket distributionBucket = new DistributionBucket(new long[]{0, 10, 50, 100});
+  DistributionBucket distributionBucket = new DistributionBucket(new long[]{0, 10_000_000, 100_000_000, 500_000_000, 1000_000_000});
 
   SLF4JReporter reporter =
       SLF4JReporter.builder().withName("metrics")
@@ -312,10 +312,9 @@ For a given array of latency distribution in milliseconds [0, 10, 100, 500, 1000
 
   Timer timer = metricRegistry.timer(metricId);    // creates a timer metric with id "latency"
 
-  long start = System.currentTimeMillis();
+  long start = Timer.start();
   // doSomething();
-  long end = System.currentTimeMillis();  
-  timer.update(end - start, tagset); // records the latency and the distribution in millisecond.
+  timer.stop(start, tagset); // records the latency and the distribution in nanoseconds.
 ```
 
 ##### DoubleValuedDistributionBucket
--- a/README.md
+++ b/README.md
@@ -289,13 +289,13 @@ There two types of distribution buckets available:
 - `DoubleValuedDistributionBucket` represented by a primitive `double` array
   
 ##### DistributionBucket
-Used to represent the distribution of an integer value. For example time spent in milliseconds or size of a messaging queue.
+Used to represent the distribution of an integer value. For example time spent in nanoseconds or size of a messaging queue.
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/ultrabrew/metrics/data/DoubleValuedDistributionBucket.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * A distribution bucket specification used for histograms of a double valued metric. A distribution
+ * bucket is an sorted array of unique numeric values. Each bucket represents a distribution range
+ * of the measurement.
+ *
+ * <p>For a given distribution array: [0, 10, 100, 500, 1000], the buckets would be like:
+ *
+ * <ul>
+ *   <li>[0,10) for values 0-9
```
manolama: use decimal values for the examples. And on line 18

core/src/main/java/io/ultrabrew/metrics/reporters/TimeWindowReporter.java
```diff
@@ -200,16 +203,39 @@ public B addHistogram(final String metricId, final DistributionBucket bucket, fi
...
-       * @param bucket distribution bucket
-       */
+    /**
+     * Add histograms to a specific metric
+     *
+     * @param metricId identifier of the metric
+     * @param bucket distribution bucket
+     * @param maxCardinality maximum cardinality of data in the histogram
+     */
+    public B addHistogram(final String metricId, final DoubleValuedDistributionBucket bucket, final int maxCardinality) {
```
manolama: Is it worth making the API a bit cleaner and having a `DistributionBucket` interface then have `DoubleValuedDistributionBucket` implement it?

manolama (commented):
Nitpick and question.

core/src/main/java/io/ultrabrew/metrics/reporters/TimeWindowReporter.java
smrutilal2: That might break backward compatibility.

smrutilal2: I have added a common interface. Please take a look

manolama (dismissed):

smrutilal2:
Can I get the second approval please @arungupta @lonemeow

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
   metricRegistry.addReporter(reporter);
 ```
 
+#### Histograms
+
+In the current implementation, clients must define the distribution buckets and associate them in the reporter with the name of the metric to be histogrammed.
+
+There two types of distribution buckets available:
+- `DistributionBucket` represented by a primitive `long` array.
+- `DoubleValuedDistributionBucket` represented by a primitive `double` array
```
arungupta, line 289: Would it be more concise to call it `DoubleDistributionBucket`? Just wondering, let's chat before changing :)

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+  
+##### DistributionBucket
+Used to represent the distribution of a integer value. For example time spent in milliseconds or size of a messaging queue.
+
+For a given distribution array: [0, 10, 100, 500, 1000], the buckets would be like:
+* [0,10) for values 0-9
```
arungupta: Since we measure time in nanos, would it be more appropriate to have an example where the buckets are sized for tracking ranges in micros / millis as very few things complete in under 100 nanos that one would measure?

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+  DoubleValuedDistributionBucket distributionBucket = new DoubleValuedDistributionBucket(new double[]{0.0, 0.5, 5.0, 10.0});
+
+  SLF4JReporter reporter =
+      SLF4JReporter.builder().withName("metrics")
+            .addHistogram(metricId, distributionBucket)    // add histogram for metric with id "auction_price"
+            .build();
+
+  String[] tagset = new String[] {"experiment", "exp1"};
+
+  GuageDouble autionPrice = metricRegistry.gaugeDouble(metricId);    // creates a guage double metric with id "auction_price"
```
arungupta: typo: `GaugeDouble`

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+
+  Timer timer = metricRegistry.timer(metricId);    // creates a timer metric with id "latency"
+
+  long start = Timer.start();
+  // doSomething();
+  timer.stop(start, tagset); // records the latency and the distribution.
+```
+
+##### DoubleValuedDistributionBucket
+Used to represent the distribution of a floating point value. For example ads auction price.
```
arungupta: Would it be more precise to say, `double-precision floating point values`?

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+
+For a given distribution array: [0.0, 0.25, 0.5, 1.0, 10.0], the buckets would be like:
+* [0.0, 0.25) for values between 0.0-0.24
```
arungupta: Is this even true? :)
Perhaps better to say `... for 0.0 <= value < 0.25` ?

smrutilal2: That's right. I will fix that.

README.md
smrutilal2: yes.

README.md
smrutilal2: Even I was considering that too. This is the similar debate that we had while naming the `GaugeDouble`.

README.md
smrutilal2: Ohh... you mean the code example.  Agreed. Fixed it to measure latencies in millis. Please, take a look.

arungupta: No, I was referring to bucket definitions. We should encourage measuring times in nanos for consistency.

smrutilal2: Okay. I have fixed that. Please take a look.

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+#### Histograms
+
+In the current implementation, clients must define the distribution buckets and associate them in the reporter with the name of the metric to be histogrammed.
+
+There two types of distribution buckets available:
+- `DistributionBucket` represented by a primitive `long` array.
+- `DoubleValuedDistributionBucket` represented by a primitive `double` array
+  
+##### DistributionBucket
+Used to represent the distribution of an integer value. For example time spent in milliseconds or size of a messaging queue.
```
arungupta: recommended correction: For example time spent in _nanoseconds_ or...

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+  DoubleValuedDistributionBucket distributionBucket = new DoubleValuedDistributionBucket(new double[]{0.0, 0.5, 5.0, 10.0});
+
+  SLF4JReporter reporter =
+      SLF4JReporter.builder().withName("metrics")
+            .addHistogram(metricId, distributionBucket)    // add histogram for metric with id "auction_price"
+            .build();
+
+  String[] tagset = new String[] {"experiment", "exp1"};
+
+  GaugeDouble autionPrice = metricRegistry.gaugeDouble(metricId);    // creates a guage double metric with id "auction_price"
```
arungupta: typo in comment: `creates a guage double`. should probably be `creates a GaugeDouble`?
also, does this actually create the gauge or gives you a reference to a gauge that had been previously created?

manolama (dismissed):

README.md
smrutilal2: It will create if not already. For the subsequent call, it will return the previously created instance.

core/src/main/java/io/ultrabrew/metrics/data/DistributionBucketIF.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A distribution bucket specification used for histograms. A distribution bucket is an sorted array
+ * of unique numeric values. Each bucket represents a distribution range of the measurement.
+ *
+ * @see DistributionBucket
+ * @see DoubleValuedDistributionBucket
+ */
+public interface DistributionBucketIF<B extends DistributionBucketIF> {
+
+  String UNDERFLOW = "underflow";
```
mmannerm, line 22: Would be good to comment how these are used. Reading the PR with a device that doesnt allow searching to see where this is used and how, but question arises is this for internal book-keeping purpose only? This is in an interface so assumption is that this is part of interface and should be documented what the expected use is.

mmannerm (dismissed):

core/src/main/java/io/ultrabrew/metrics/data/DistributionBucketIF.java
smrutilal2: Added the doc. Please, take a look

mmannerm (dismissed):

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+
+There two types of distribution buckets available:
+- `DistributionBucket` represented by a primitive `long` array.
+- `DoubleValuedDistributionBucket` represented by a primitive `double` array
+  
+##### DistributionBucket
+Used to represent the distribution of an integer value. For example time spent in nanoseconds or size of a messaging queue.
+
+For a given array of latency distribution in nanoseconds [0, 10_000_000, 100_000_000, 500_000_000, 1000_000_000], the buckets would be like:
+* [0,10_000_000) for values 0 - 9_999_999
```
arungupta: [0, 10_000_000) for 0 <= value < 10ms

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+      SLF4JReporter.builder().withName("metrics")
+            .addHistogram(metricId, distributionBucket)    // add histogram for metric with id "auction_price"
+            .build();
+
+  String[] tagset = new String[] {"experiment", "exp1"};
+
+  GaugeDouble autionPrice = metricRegistry.gaugeDouble(metricId);    // creates a gauge double metric with id "auction_price"
+
+  double price = getAuctionPrice();
+  autionPrice.set(price, tagset1); // records the auction_price and the distribution.
```
arungupta: typo `auctionPrice`

`auctionPrice.set(getAuctionPrice(), tagset)`

README.md
```diff
@@ -280,6 +280,72 @@ SLF4J Logger.
...
+* [0.0, 0.25) for 0.0 <= value < 0.25
+* [0.25, 0.5) for 0.25 <= value < 0.5
+* [0.5, 1.0) for 0.5 <= value < 1.0
+* [1.0, 10.0) for 1.0 <= value < 10.0
+* overflow  for values  >= 10.0
+* underflow for values  < 0.0
+
+```Java
+  String metricId = "auction_price";
+  DoubleValuedDistributionBucket distributionBucket = new DoubleValuedDistributionBucket(new double[]{0.0, 0.5, 5.0, 10.0});
```
arungupta: might be better to use the same buckets as the example on line 323 for better continuity.

smrutilal2:
I have addressed your comments. Please take a look @arungupta

arungupta (approved):

manolama (approved):

PR #49, opened 2021-07-20
Fix JVM CodeHeap memory pool collection

rafan:
Since Java 9, code heap was split into three different pools. Their names containing single-quote (`'`) which is not acceptable by my metric collector. To be honest, I think most collector does not like single-quote. Therefore, it makes sense to sanitize the names (like it did with the space character) in the Ultrabrew Metrics library.


manolama (approved):
LGTM, thanks!

lonemeow:
This will, of course, change the metric names for anyone who is currently using them on Java 9+ successfully with a collector that has no issues with such characters.

manolama:
True, so we _could_ change it in the reporters possibly.

lonemeow:
Arguably, reporters should fixup characters they can not deal with, since the core code can't know about every possible reporters' limitations.

manolama:
@rafan mind putting this in our internal repo instead please?

rafan:
Sure, I will do it for the internal reporter.

PR #51, opened 2021-10-14
add jvm.cpu.usage metric

mshauneu:

### DIFF
--- a/core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
+++ b/core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
@@ -12,8 +12,13 @@ import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryUsage;
+import java.lang.management.OperatingSystemMXBean;
 import java.lang.management.ThreadMXBean;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
@@ -50,6 +55,8 @@ import java.util.concurrent.TimeUnit;
  * <dd>Total number of garbage collections for this collector.</dd>
  * <dt>jvm.gc.&lt;collectorname&gt;.time</dt>
  * <dd>Approximate total time used by this collector in milliseconds.</dd>
+ * <dt>jvm.cpu.usage</dt>
+ * <dd>CPU usage in percents by JVM.</dd>
  * </dl>
  *
  * @see ClassLoadingMXBean
@@ -58,6 +65,7 @@ import java.util.concurrent.TimeUnit;
  * @see MemoryPoolMXBean
  * @see BufferPoolMXBean
  * @see GarbageCollectorMXBean
+ * @see OperatingSystemMXBean
  */
 public class JvmStatisticsCollector {
 
@@ -66,6 +74,18 @@ public class JvmStatisticsCollector {
   private final Semaphore exitRequest = new Semaphore(0);
   private Thread reportingThread = null;
 
+
+  private static final List<String> OPERATING_SYSTEM_BEAN_CLASS_NAMES = Arrays.asList(
+    "com.ibm.lang.management.OperatingSystemMXBean", // J9
+    "com.sun.management.OperatingSystemMXBean" // HotSpot
+  );
+
+  private final OperatingSystemMXBean operatingSystemBean;
+  private final Class<?> operatingSystemBeanClass;
+
+  private final Method processCpuUsage;
+
+
   /**
    * Create a JvmStatisticsCollector that emits statistics to the given {@link MetricRegistry}.
    *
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
```diff
@@ -12,8 +12,13 @@
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryUsage;
+import java.lang.management.OperatingSystemMXBean;
 import java.lang.management.ThreadMXBean;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
```
smrutilal2: Reflection would be red flag.

smrutilal2: This library is very performance sensitive. If we can't avoid it then, we might have to make it optional. Let's see what others have to say about it.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
```diff
@@ -66,13 +74,29 @@
...
   /**
    * Create a JvmStatisticsCollector that emits statistics to the given {@link MetricRegistry}.
    *
    * @param registry registry to report statistics to
    */
   public JvmStatisticsCollector(final MetricRegistry registry) {
     this.registry = registry;
+
+    this.operatingSystemBean = ManagementFactory.getOperatingSystemMXBean();
+    this.operatingSystemBeanClass = getFirstClassFound(OPERATING_SYSTEM_BEAN_CLASS_NAMES);
```
lonemeow: Why do we need this instead of using `this.operatingSystemBean.getClass()`?

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
```diff
@@ -166,5 +190,43 @@ private void collect() {
       setGauge(prefix + ".count", collectorBean.getCollectionCount());
       setGauge(prefix + ".time", collectorBean.getCollectionTime());
     }
+
+    if (processCpuUsage != null) {
+      setGauge("jvm.cpu.usage", invoke(processCpuUsage));
+    }
+  }
+
+  private long invoke(Method method) {
```
lonemeow: The name is overly generic given the implementation is very specific invocation.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
```diff
@@ -166,5 +190,43 @@ private void collect() {
...
+    }
   }
+
+  private Method detectMethod(String name) {
+    if (operatingSystemBeanClass == null) {
+      return null;
+    }
+    try {
+      operatingSystemBeanClass.cast(operatingSystemBean);
+      return operatingSystemBeanClass.getDeclaredMethod(name);
```
lonemeow: This should probably also ensure the method has correct signature, not just name alone.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
```diff
@@ -66,13 +74,29 @@
...
+
+  private static final List<String> OPERATING_SYSTEM_BEAN_CLASS_NAMES = Arrays.asList(
+    "com.ibm.lang.management.OperatingSystemMXBean", // J9
+    "com.sun.management.OperatingSystemMXBean" // HotSpot
+  );
+
+  private final OperatingSystemMXBean operatingSystemBean;
+  private final Class<?> operatingSystemBeanClass;
+
+  private final Method processCpuUsage;
```
lonemeow: Using `MethodHandle` for dynamic invocation would be preferable, as it pushes all the checks to initialization time instead of invocation time, which both improves performance and results in better error behavior.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
mshauneu: `this. operatingSystemBean` has type `OperatingSystemMXBean` that is interface.  Method`getFirstClassFound` tries to find a class that implements `OperatingSystemMXBean` interface.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
mshauneu: migrated to `MethodHandle`. No reflection anymore.

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
mshauneu: migrated to MethodHandle

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
mshauneu: migrated to MethodHandle

core/src/main/java/io/ultrabrew/metrics/JvmStatisticsCollector.java
mshauneu: I chose generic name for all 3 methods [`getFirstClassFound`, `detectMethod` and `invoke` because they could be used for other [stats,](https://docs.oracle.com/javase/8/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html) if required, e.g:

```java
this.getProcessCpuTime = detectMethod(operatingSystemBeanClass, "getProcessCpuTime");
...
setGauge("jvm.cpu.time", invoke(getProcessCpuTime));
```

mshauneu: invoke gone during migration to `MethodHandle`

